PR #9, opened 2019-05-03
Added coverage report to the repository

sgnn7:
We now have visibility into what code is not being adequately tested or
checked. Any issues found during the enabling of this feature has been
integrated too so that we can have the highest targets possible.

Build: https://jenkins.conjur.net/job/conjurinc--conjur-api-python3/job/coverage/

### DIFF
--- a/test/test_api.py
+++ b/test/test_api.py
@@ -318,7 +318,7 @@ class ApiTest(unittest.TestCase):
 
     @patch('conjur_api_python3.api.invoke_endpoint', \
            return_value=MockClientResponse(content=json.dumps(MOCK_RESOURCE_LIST)))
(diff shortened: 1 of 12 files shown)


test/test_api.py
```diff
@@ -318,7 +318,7 @@ def mock_auth():
 
     @patch('conjur_api_python3.api.invoke_endpoint', \
            return_value=MockClientResponse(content=json.dumps(MOCK_RESOURCE_LIST)))
-    def test_get_variables_invokes_http_client_correctly(self, mock_http_client):
+    def test_get_resoruces_invokes_http_client_correctly(self, mock_http_client):
```
izgeri:
```suggestion
    def test_get_resources_invokes_http_client_correctly(self, mock_http_client):
```

PR #10, opened 2019-05-03
Add policy change output support

sgnn7:
We now properly output results from apply and replace policy invocations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-# VIM swapfiles
-*.sw[po]
-
 # venv files
 venv/
 
--- a/test/test_cli_integration.py
+++ b/test/test_cli_integration.py
@@ -71,6 +71,28 @@ class CliIntegrationTest(unittest.TestCase): # pragma: no cover
         return invoke_cli(self, self.cli_auth_params,
             ['policy', 'replace', 'root', policy_path])
 
+    def replace_policy_from_string(self, policy):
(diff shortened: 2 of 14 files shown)


test/test_cli_integration.py
```diff
@@ -71,6 +71,28 @@ def replace_policy(self, policy_path):
...
+        with tempfile.NamedTemporaryFile() as temp_policy_file:
+            temp_policy_file.write(policy.encode('utf-8'))
+            temp_policy_file.flush()
+
+            # Run the new apply that should not result in newly created roles
+            output = self.replace_policy(temp_policy_file.name)
+
+        return output
+
+    def apply_policy_from_string(self, policy):
```
dividedmind: Step in the right direction, though I think this could be DRYed up even more.

sgnn7: Yeah this needs some DRYing up for sure as the tests are getting bulkier. I might defer this to next piece of work on this file so that we can get this into PyPI.

sgnn7:
@dividedmind CR ready again

.gitignore
```diff
@@ -1,6 +1,9 @@
 # VIM swapfiles
 *.sw[po]
 
+# OSX
+**/.DS_Store
```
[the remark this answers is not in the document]
infamousjoeg: TIL!

PR #15, opened 2019-07-12
14 - Fix conjurrc empty value overrides

### LINKED ISSUE #14: conjur-cli is failing to read .conjurrc correctly and is throwing error "Account cannot be empty"
When using the conjur-cli with .conjurrc the url, account and ca_bundle variables are set to None resulting in a "Account cannot be empty" exception. The reason for this is how we update the config dictionary here.
https://github.com/cyberark/conjur-api-python3/blob/master/conjur/client.py#L72-L74

To fix this issue we must filter the config of all values that are None:

```python
filtered_config = {k:v for k, v in config.items() if v is not None}
on_disk_config.update(filtered_config)
config = on_disk_config
```

sgnn7:
Fixes #14 

We now properly override `None` values in `Client` initialization with the ones from conjurrc.

CC: @AndrewCopeland

### DIFF
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -17,6 +17,8 @@ class MockApiConfig(object):
         'url': 'apiconfigurl',
         'account': 'apiconfigaccount',
         'ca_bundle': 'apiconfigcabundle',
+        'login_id': 'apiconfigloginid',
+        'api_key': 'apiconfigapikey',
     }
 
     def __iter__(self):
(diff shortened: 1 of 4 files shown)


test/test_client.py
```diff
@@ -143,6 +143,18 @@ def verify_init_args(api_instance, **kwargs):
         Client(api_class=MockApi, api_config_class=MockApiConfig, url='http://foo',
                account='myacct', login_id='mylogin', ca_bundle='mybundle')
 
+    def test_client_does_not_override_apiconfig_values_with_empty_values(self):
+        class MockApi(MockApiHelper):
+            def verify_init_args(api_instance, **kwargs):
+                api_instance.verify_apiconfig_dict_in(self, **kwargs)
+                self.assertEqual(kwargs['account'], 'apiconfigaccount')
+                self.assertEqual(kwargs['url'], 'apiconfigurl')
+                self.assertEqual(kwargs['ca_bundle'], 'apiconfigcabundle')
```
izgeri: should the `login_id` also be checked here?

sgnn7: Good catch. Fixing.

PR #16, opened 2019-07-30
Update README.md include From PyPI

AndrewCopeland:


sgnn7:
@AndrewCopeland Thanks!

PR #18, opened 2019-08-06
Submit CodeClimate Coverage

hughsaunders:
* submit coverage results to CodeClimate
* Add gitleaks.toml config file

Related: conjurinc/ops#233

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,224 @@
+title = "conjur-api-python3 Gitleaks Configuration"
+
+# This is a sample config file for gitleaks. You can configure gitleaks what to search for and what to whitelist.
+# The output you are seeing here is the default gitleaks config. If GITLEAKS_CONFIG environment variable
+# is set, gitleaks will load configurations from that path. If option --config-path is set, gitleaks will load
+# configurations from that path. Gitleaks does not whitelist anything by default.
+# - https://www.ndss-symposium.org/wp-content/uploads/2019/02/ndss2019_04B-3_Meli_paper.pdf
+# - https://github.com/dxa4481/truffleHogRegexes/blob/master/truffleHogRegexes/regexes.json
+[[rules]]
+description = "AWS Client ID"
+regex = '''(A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16}'''
+tags = ["key", "AWS"]
+
+[[rules]]
+description = "AWS Secret Key"
+regex = '''(?i)aws(.{0,20})?(?-i)['\"][0-9a-zA-Z\/+]{40}['\"]'''
+tags = ["key", "AWS"]
+
+[[rules]]
+description = "AWS MWS key"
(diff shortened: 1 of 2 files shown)


.gitleaks.toml
```diff
@@ -0,0 +1,224 @@
...
+regex = '''(?i)twilio(.{0,20})?['\"][0-9a-f]{32}['\"]'''
+tags = ["key", "twilio"]
+
+[whitelist]
+files = [
+  "(.*?)(jpg|gif|doc|pdf|bin)$",
+  "test/test_config/https/conjur.key",
+  ".gitleaks.toml"
+]
+#regexes = [
```
sgnn7: We should probably remove the commented out examples

.gitleaks.toml
```diff
@@ -0,0 +1,224 @@
...
+[[rules]]
+description = "PGP"
+regex = '''-----BEGIN PGP PRIVATE KEY BLOCK-----'''
+tags = ["key", "PGP"]
+
+[[rules]]
+description = "Facebook Secret Key"
+regex = '''(?i)(facebook|fb)(.{0,20})?(?-i)['\"][0-9a-f]{32}['\"]'''
+tags = ["key", "Facebook"]
+
```
sgnn7: I know they're there to prevent key leaks but I don't see us ever committing the following keys to the repo so they should probably be removed:
- FB
- Twitter
- Linkedin
- Mailchimp
- Mailgun
- Paypal
- Stripe
- Square

Jenkinsfile
```diff
@@ -24,6 +24,7 @@ pipeline {
         always {
           junit 'output/**/*.xml'
           cobertura autoUpdateHealth: true, autoUpdateStability: true, coberturaReportFile: 'coverage.xml', conditionalCoverageTargets: '100, 0, 0', failUnhealthy: true, failUnstable: false, lineCoverageTargets: '99, 0, 0', maxNumberOfBuilds: 0, methodCoverageTargets: '100, 0, 0', onlyStable: false, sourceEncoding: 'ASCII', zoomCoverageChart: false
+          ccCoverage("coverage.py")
```
sgnn7: We probably need a config to ensure that we don't reduce our line coverage from codeclimate perspective too

.gitleaks.toml
hughsaunders: No harm in having a comprehensive config though

.gitleaks.toml
hughsaunders: I disagree, I think they are useful for showing whats possible next time someone needs to change the configuration.

Jenkinsfile
hughsaunders: I set 75% diff coverage in the UI, as it's not possible (AFAIK) to configure coverage threshold in repo. I've also opened a ticket with codeclimate about that.

sgnn7: 👍

.gitleaks.toml
sgnn7: I think there is harm: the longer processing time for each push and failure to use [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) to reduce code complexity where that complexity doesn't yield benefits. I think my problem is that I can't think of a single theoretical future with a key of the type listed being included in the repo .

Anyways, I'm cool with this staying as-is for now given the newer hooks we developed for pushes that run decently fast but I'm not sure they won't be trimmed if the scanning gets too slow.

PR #49, opened 2020-04-15
Builds fail only on fixable HIGH or CRITICAL vulns

andytinkham:
Changes the trivy scan to do 2 reports - one of just fixable potential vulnerabilites and one with all detected potential vulnerabilities (whether a fix is available or not). Builds will only fail when there is a fix available. I also bumped the build failure threshold so builds will fail on fixable HIGH or CRITICAL vulnerabilities instead of just CRITICAL, as there should be no fixable HIGHs currently. 
Signed-off-by: Andy Tinkham <andy.tinkham@cyberark.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -73,8 +73,18 @@ pipeline {
     }
 
     stage('Scan Docker image') {
-      steps{
-        scanAndReport("conjur-python-cli:latest", "CRITICAL")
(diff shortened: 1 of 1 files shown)


andytinkham:
WIP because integration tests are failing prior to these changes. Need to investigate further.

PR #61, opened 2020-09-03
Make CONTRIBUTING refer to contrib license guide

izgeri:
### What does this PR do?
Depends on cyberark/community#74

Connected to cyberark/community#1

#### Change log

#### Test coverage

#### Documentation
- [x] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,11 @@
 # Contributing
 
 For general contribution and community guidelines, please see the [community repo](https://github.com/cyberark/community).
+In particular, before contributing please review our [contributor licensing guide](https://github.com/cyberark/community/blob/master/CONTRIBUTING.md#when-the-repo-does-not-include-the-cla)
+to ensure your contribution is compliant with our contributor license agreements.
 
 ## Table of Contents
 
-- [Legal mumbo-jumbo](#legal-mumbo-jumbo)
 - [Building](#building)
 - [Development](#development)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,10 +1,11 @@
 # Contributing
 
 For general contribution and community guidelines, please see the [community repo](https://github.com/cyberark/community).
+In particular, before contributing please review our [contributor licensing guide](https://github.com/cyberark/community/blob/master/CONTRIBUTING.md#when-the-repo-does-not-include-the-cla)
```
jtuttle, line 4: The piece of the URL after the `#` that specifies a header doesn't seem to do anything. I'm guessing it refers to a header that got removed? We should probably just link to the file without any header specified since they're more prone to changing than the file itself.

jtuttle (changes requested):

CONTRIBUTING.md
izgeri: it's actually linking to a header I'm adding - that's why this PR depends on cyberark/community#74

there's a lot in that file, so I'd prefer to leave the header to point to the specific section if I can. if it helps convince you, I'm making this update across a bunch of projects to ensure links continue to work :) I'm committed to doing the same if I change the links in the future!

jtuttle (approved):

PR #65, opened 2020-10-28
Upgrade urllib to 1.25.9+

orenbm:
urllib3 before 1.25.9 allows CRLF injection if the attacker controls the HTTP request method,
as demonstrated by inserting CR and LF control characters in the first argument of putrequest().
NOTE: this is similar to CVE-2020-26116.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,5 +7,5 @@ PyYAML>=3.13
 requests>=2.21.0
 twine>=1.13.0
 
-# https://github.com/kennethreitz/requests/issues/5065
-urllib3>=1.24.2,<1.25
+# https://nvd.nist.gov/vuln/detail/CVE-2020-26137
+urllib3>=1.25.9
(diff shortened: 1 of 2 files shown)


sigalsax:
I doubt it but do we know if this has cascading breaking changes? Did you validate that the CLI still works with this version bump?

sgnn7:
@sigalsax This repo has really good tests so I have confidence that if they pass this change is good. If there's a problem, it's an extremely unlikely one.

sgnn7 (dismissed):
@orenbm LGTM!

sgnn7:
@orenbm Can you also change this section too? https://github.com/cyberark/conjur-api-python3/blob/master/setup.py#L45

orenbm:
@sgnn7 can you re-review please?

sgnn7 (approved):
@orenbm LGTM!

PR #96, opened 2020-11-18
Add variable CLI command design

sigalsax:
### What does this PR do?

Adds design for variable CLI command

### DIFF
--- a/design/variable_command_design.md
+++ b/design/variable_command_design.md
@@ -1,0 +1,149 @@
+# Dockerless CLI - `variable` action
+
+This document is the design document for the variable CLI command for the Dockerless project. The functionality for the variable CLI command already exists so all that is left is the following:
+
+1. Ensure UTs and integration tests exist and are GA quality
+3. Write proper documentation in `README.md` of the repository
+
+For the full UX design for the list command see [here](https://ljfz3b.axshare.com/#id=yokln4&p=conjur_main_help&g=1).
+
+### UX mapping
+
+We use a third-party module called [argparse](https://docs.python.org/3/library/argparse.html) which gives us limited control over the types of errors we can return. Therefore, at this time we will not surgicially change the errors and leave them as is. The following is a mapping of the current flow for all possible permutations of the variable CLI command and their expected outputs/behavior. 
+
+*Input for variable is neither get/set*
+
+```bash
+$ conjur-cli variable boo
+
+## Current
+usage: conjur-cli variable [-h] {get,set} ...
+conjur-cli variable: error: argument action: invalid choice: 'boo' (choose from 'get', 'set')
+
+## Should be
+conjur-cli variable: error: argument action: invalid choice: 'boo' (choose from 'get', 'set')
+
+<Variable command help screen>
+```
+
+*One of the batch secrets doesn’t have value/doesn’t exist*
+
+```bash
+$ conjur-cli variable get secrets/test_secret foo/bar
+
+## Current
+requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://localhost/secrets/myorg/variable/secrets%2Ftest_secret%2Cfoot%2Fbar
+
+## Should be
+Error: Data either not found or variable value does not exist
+
+<Variable command help screen>
+```
+
+*Set secrets*
+
+```bash
+$ conjur-cli variable set foo/bar mysecret
+
+## Current
+Value set: 'foo/bar'
+
+## Should be
+Success! Data set to: 'foo/bar'
+```
+
+*Get secrets*
+
+```bash
+$ conjur-cli variable get foo/bar
+mysecret%
+```
+
+*Get a secret without inputting which secret to get*
+
+```bash
+$ conjur-cli variable get __
+
+## Current 
+usage: conjur-cli variable get [-h] variable_id [variable_id ...]
+conjur-cli variable get: error: the following arguments are required: variable_id
+
+## Should be
+conjur-cli variable get: error: the following arguments are required: variable_id
+
+<Variable command help screen>
+```
+
+*Get a secret without inputting which variable_id / its value*
+
+```bash
+$ conjur-cli variable set __ __
+
+## Current
+usage: conjur-cli variable set [-h] variable_id value
+conjur-cli variable set: error: the following arguments are required: variable_id, value
+
+## Should be
+conjur-cli variable set: error: the following arguments are required: variable_id, value
+
+<Variable command help screen>
+```
+
+*Get secret without inputting the variable_id’s value*
+
+```bash
+$ conjur-cli variable set foo/bar __
+
+## Current
+usage: conjur-cli variable set [-h] variable_id value
+conjur-cli variable set: error: the following arguments are required: value
+
+## Should be
+conjur-cli variable set: error: the following arguments are required: value
+
+<Variable command help screen>
+```
+
+*Get secrets with more inputs than excepted*
+
+```bash
+$ conjur-cli variable set foo/bar mysecret extraargument
+
+## Current
+usage: conjur-cli [-h] [-v] [-l URL] [-a ACCOUNT] [-c CA_BUNDLE] [--insecure] [-u LOGIN_ID] [-k API_KEY] [-p PASSWORD] [-d]
+         [--verbose]
+         {list,variable,policy} ...
+conjur-cli: error: unrecognized arguments: extraargument
+
+## Should be:
+conjur-cli: error: unrecognized arguments: extraargument
+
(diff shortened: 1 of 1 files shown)


design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
...
+
+2. Should we return error codes? 404 if not found on secrets?
+
+### Delivery plan
+
+1. Ensure UTs and integration tests exist and are GA quality (3 days)
+   1. Sit with Elad to map out current tests and what is missing
+   2. Implement missing tests (if required)
+2. Improve error handling experience (3 days)
+   1. ***Blocker*** - need to get full UX outlines from Sharon + Inbal
```
InbalZilberman, line 177: I think we agreed on \" and spaces
conjur variable get "var1" "var2"
OK?
Ill update the Epic also

design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
...
+### Open questions
+
+1. If extra arguments are provided, what should be the experience? See second to last flow.
+
+2. Should we return error codes? 404 if not found on secrets?
+
+### Delivery plan
+
+1. Ensure UTs and integration tests exist and are GA quality (3 days)
+   1. Sit with Elad to map out current tests and what is missing
```
InbalZilberman, line 174: I think it is best to write QA & PO  :)

design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
...
+$ conjur-cli variable set foo/bar mysecret sfdfs
+
+## Current
+usage: conjur-cli [-h] [-v] [-l URL] [-a ACCOUNT] [-c CA_BUNDLE] [--insecure] [-u LOGIN_ID] [-k API_KEY] [-p PASSWORD] [-d]
+         [--verbose]
+         {list,variable,policy} ...
+conjur-cli: error: unrecognized arguments: sfdfs
+
+## Should be:
+Should the last input be ignored? Should it take only the arguments it needs (first 2) as we do in Ruby CLI?
```
InbalZilberman, line 148: yes. Ignored

design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
...
+mysecret%
+```
+
+*Get a secret without inputting which secret to get*
+
+```bash
+$ conjur-cli variable get __
+
+## Current 
+usage: conjur-cli variable get [-h] variable_id [variable_id ...]
```
InbalZilberman, line 98: usage: conjur-cli variable get [-h] **"**variable_id**"** [**"**variable_id**"** ...]

design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Delivery plan
+
+1. Ensure UTs and integration tests exist and are GA quality (3 days)
+   1. Sit with Elad to map out current tests and what is missing
+   2. Implement missing tests (if required)
+2. Improve error handling experience (3 days)
+   1. ***Blocker*** - need to get full UX outlines from Sharon + Inbal
+3. Write and get approval on UX logs (2 days)
+4. Write proper documentation in README.md of the repository (2 days)
```
InbalZilberman, line 179: 5. Write proper documentation in confluence and open doc ticket

design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
...
+*Get a secret without inputting which secret to get*
+
+```bash
+$ conjur-cli variable get __
+
+## Current 
+usage: conjur-cli variable get [-h] variable_id [variable_id ...]
+
+## Should be
+Error: Command 'variable' requires the following arguments: variable_id
```
sgnn7, line 101: This command accepts one-or-more variable IDs so the description needs to describe that wording. I do think that the original version of the help here is much more helpful for a CLI user.

sgnn7 (commented):
@sigalsax Left a few comments

design/variable_command_design.md
sigalsax: yes but I am talking about the actual help screen, info, and error statements here

design/variable_command_design.md
sigalsax: this the current flow and I am suggesting below:
`Error: Command 'variable' requires the following arguments: variable_id`

design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
+# Dockerless CLI - variable
+
+This document is the design document for the variable CLI command for the Dockerless project. The functionality for the variable CLI command already exists so all that is left is the following:
```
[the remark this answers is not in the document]
sigalsax: The line or the sentence?

design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
...
+*Set secrets*
+
+```bash
+$ conjur-cli variable set foo/bar mysecret
+
+## Current
+Value set: 'foo/bar'
+
+## Should be
+Success! Data written to: 'foo/bar'
```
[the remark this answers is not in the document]
sigalsax: Out of curiousity, what makes this non-machine readable? The `''` and the `!`?

design/variable_command_design.md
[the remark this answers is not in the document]
sgnn7: Line itself. This is easily over 100 characters so it needs breaking down into multiple lines..

design/variable_command_design.md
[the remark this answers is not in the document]
sgnn7: If you try to give me a way to parse the variable name from each of those outputs, you will see that the suggested use case requires careful counting vs otherwise. Both can be extracted the same way but the original output is less prone to error during script development.

design/variable_command_design.md
```diff
@@ -0,0 +1,181 @@
...
+
+```bash
+$ conjur-cli variable boo
+
+## Current
+usage: conjur-cli variable [-h] {get,set} ...
+conjur-cli variable: error: argument action: invalid choice: 'boo' (choose from 'get', 'set')
+
+## Should be
+Error: Command 'variable' requires a subcommand get, set
```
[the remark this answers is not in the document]
sigalsax: So it seems that argparse is a wrapper for our error handling and help screens giving us the form
```
usage: myprog.py [-h] xyz

positional arguments:
  xyz   ...

optional arguments:
  -h, --help  ..
```
Requiring us to switch this out for something else, or build one ourself (please no), or create a wrapper for catching these methods and returning them in the format suggested above?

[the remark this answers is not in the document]
sigalsax: After some research I found [this](https://stackoverflow.com/questions/35847084/customize-argparse-help-message) that could help us get around the format.

TL;DR, we could do the following:
1. Suppress the help given to us and providing our own
2. Override optional arguments/positional (_optional_argument, _position_argument) but these are internal attributes and may lead to breaking changes if the module is updated in this area
3. Provide a wrapper for the usage class as explained in the link

These are some idea but requires more research. What do you think?

design/variable_command_design.md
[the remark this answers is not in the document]
sigalsax: why careful counting if we split at the spaces and take the last index. Can you explain a bit more about what you mean by this or why current output makes it less error prone?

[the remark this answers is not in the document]
sigalsax: As we spoke about that the writing variables isn’t critical as is reading of variable be machine-readable so it’s ok

design/variable_command_design.md
```diff
@@ -0,0 +1,179 @@
...
+*Set secrets*
+
+```bash
+$ conjur-cli variable set foo/bar mysecret
+
+## Current
+Value set: 'foo/bar'
+
+## Should be
+Success! Data written to: 'foo/bar'
```
sigalsax, line 78: @sgnn7 what about `Success! Data set to: ''`?

design/variable_command_design.md
sigalsax: This was a mistake, the current error is:
```
usage: conjur-cli variable get [-h] variable_id [variable_id ...]
conjur-cli variable get: error: the following arguments are required: variable_id
```

design/variable_command_design.md
sgnn7: Sounds good to me

design/variable_command_design.md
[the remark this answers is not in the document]
sigalsax: I use [Typora](https://typora.io/) when I write MD so I can't measure how it looks in raw markdown. Is this critical because MDs are rendered in GH and won't be an excessive long line as we see when in raw

InbalZilberman (approved):
Great work! thank you

eladkug (approved):
LGTM

PR #97, opened 2020-11-18
Add list command design

sigalsax:
### What does this PR do?

Add list CLI command design

### DIFF
--- a/design/list_command_design.md
+++ b/design/list_command_design.md
@@ -1,0 +1,267 @@
+# Dockerless CLI - `list` action
+
+This document is the design document for the *list* CLI command for the Dockerless project. The main functionality for the list CLI command has been implemented. Now all that is left is to implement the following flags:
+
+1. `-i, --inspect`
+2. `-k, --kind`
+3. `-l, --limit`
+5. `-r, --role`
+6. `-s, --search`
+6. `-o, --offset` (low in priority)
+
+`offset` is at a lower priority because the current Ruby CLI does not implement this flag.
+
+In additional to implementing the above flags, we will need to also address the following:
+
+1. Ensure UTs and integration tests exist and are GA quality
+3. Write proper documentation in `README.md` of the repository
+
+For the full UX design for the list command see [here](https://ljfz3b.axshare.com/#id=yokln4&p=conjur_main_help&g=1).
+
+### UX mapping
+
+We use a third-party module called [argparse](https://docs.python.org/3/library/argparse.html) which gives us limited control over the types of errors we can return. Therefore, at this time we will not surgicially change the errors and leave them as is. The following is a mapping of the current flow for all possible permutations of the *list* CLI command and their expected outputs/behavior. 
+
+*When no values were found*
+
+```bash
+$ conjur-cli list
+
+## Current
+[]
+
+## Should be
+No results found
+```
+
+*List with more inputs than excepted*
+
+```bash
+$ conjur-cli list unknown
+
+## Current
+usage: conjur-cli [-h] [-v] [-l URL] [-a ACCOUNT] [-c CA_BUNDLE] [--insecure] [-u LOGIN_ID] [-k API_KEY] [-p PASSWORD] [-d]
+         [--verbose]
+         {list,variable,policy} ...
+conjur-cli: error: unrecognized arguments: boo
+
+## Should return an error
+```
+
+*Mispell list command*
+
+```bash
+$ conjur-cli listboo
+
+## Current
+usage: conjur-cli [-h] [-v] [-l URL] [-a ACCOUNT] [-c CA_BUNDLE] [--insecure] [-u LOGIN_ID] [-k API_KEY] [-p PASSWORD] [-d]
+                  [--verbose]
+                  {list,variable,policy} ...
+conjur-cli: error: argument resource: invalid choice: 'listboo' (choose from 'list', 'variable', 'policy')
+
+## Should be:
+conjur-cli: error: argument resource: invalid choice: 'listboo' (choose from 'list', 'variable', 'policy')
+
+<Main help screen>
+```
+
+*Help flag*
+
+```bash
+$ conjur-cli list --help
+
+## Current
+usage: conjur-cli list [-h]
+
+optional arguments:
+  -h, --help  show this help message and exit
+  
+## Should be
+<List help screen>
+```
+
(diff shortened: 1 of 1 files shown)


design/list_command_design.md
```diff
@@ -0,0 +1,289 @@
...
+----
+--[no-]inspect       Show full object information
+
+-k, --kind=arg       Filter by kind (default: none)
+
+-l, --limit=arg      Maximum number of records to return (default: none)
+
+-o, --offset=arg     Offset to start from (default: none)
+
+--role=ROLE          Role to act as. By default, the current logged-in role is used. (default: none)
```
eladkug, line 45: why we dont define here shortcut as well ? 
and its need to be declared better how the role suppose to look like (you need to add the org before the role)

InbalZilberman: @eladkug you are right
shortcut should be -r

eladkug: @InbalZilberman -r may be a problem since its deprecated for python (and still exist in ruby cli)
`-r, --[no-]raw-annotations - Show annotations in 'raw' format`

design/list_command_design.md
```diff
@@ -0,0 +1,275 @@
...
+  -d, --debug				 Enable debugging output
+
+       
+Commands
+----
+--[no-]inspect       Show full object information
+
+-k, --kind=arg       Filter by kind (default: none)
+
+-l, --limit=arg      Maximum number of records to return (default: none)
```
sgnn7, line 40: I just realized that `--limit` isn't really all that useful without a complementary `--offset` flag too

design/list_command_design.md
```diff
@@ -0,0 +1,275 @@
+# Dockerless CLI - `list` action
+
+This document is the design document for the list CLI command for the Dockerless project. The main functionality for the list CLI command has been implemented. Now all that is left is to implement the following flags:
```
sgnn7, line 3: Line-too-long

design/list_command_design.md
sigalsax: For clarity purposes? So if the user has 100 entries they can start in the middle let's say?

sigalsax: FYI @InbalZilberman

InbalZilberman: I totally agree with @sgnn7 limit and offset are a pair
Does the REST API support offset?
If not I rather we support what we can and start to write the gaps in an orderly way so we can take them as we go along.

sigalsax: Yes the REST API supports offset. Is this back in scope?

design/list_command_design.md
InbalZilberman: lets go with -r for role and when we add raw-annotations i think we will add it as -a

sigalsax: @InbalZilberman why do we need the `=` between each key and value. From what I have seen it is usually `-<flag> <flag-value>`

design/list_command_design.md
InbalZilberman: @sigalsax I think both ways will work but sure lets make it simpler
please notice that in Ruby only these option mention with '=' are you sure there is no technical reason for it?

sigalsax: did you mean to comment this on the above comment thread? https://github.com/cyberark/conjur-api-python3/pull/97#discussion_r528612619

InbalZilberman: So yes lets do it in lower priority though

design/list_command_design.md
```diff
@@ -0,0 +1,266 @@
...
+  "myorg:policy:root",
+  "myorg:policy:secrets",
+  "myorg:policy:users"
+]
+```
+
+*`--limit=-1` (negative number or 0) gives us the full list.* In the Ruby CLI, this is an 500 Internal Server Error
+
+```bash
+$ conjur list --limit=-1
```
eladkug, line 190: I think an error should be the result.
As a user i want to know that i add wrong input (for example in automation if i will do by mistake -1), the user may think its working when he have 1 item (it will show all the list) but then when we will have another item he will not understand why he get two items instead of 1.

sharonr78: Agree with @eladkug, as showing the full list will draw too much of the user's attention from the actual error.

InbalZilberman: @sigalsax what happens in the REST API?

InbalZilberman: We can go with @eladkug suggestion and try it out in our UX tests

sigalsax: @InbalZilberman the REST API returns a `HTTP/1.1 500 Internal Server Error`

design/list_command_design.md
```diff
@@ -0,0 +1,272 @@
...
+$ conjur-cli listboo
+
+## Current
+usage: conjur-cli [-h] [-v] [-l URL] [-a ACCOUNT] [-c CA_BUNDLE] [--insecure] [-u LOGIN_ID] [-k API_KEY] [-p PASSWORD] [-d]
+                  [--verbose]
+                  {list,variable,policy} ...
+conjur-cli: error: argument resource: invalid choice: 'listl' (choose from 'list', 'variable', 'policy')
+
+## Should be:
+conjur-cli: error: argument resource: invalid choice: 'listn' (choose from 'list', 'variable', 'policy')
```
eladkug, line 63: you meant invalid choice: 'listboo', right?

design/list_command_design.md
```diff
@@ -0,0 +1,272 @@
...
+  "myorg:policy:secrets",
+  "myorg:policy:users"
+]
+```
+
+*`--limit=-1` (negative number or 0) gives us the full list.* In the Ruby CLI, this is an 500 Internal Server Error
+
+```bash
+$ conjur list --limit=-1
+[
```
eladkug, line 198: Change the example , you suppose to get error

design/list_command_design.md
sigalsax: 💯

design/list_command_design.md
sigalsax: Do you think it would be a 500 internal server error? Or simply no results found?

eladkug: I think the user need to get an error on argument that it have to be positive value, greater than 0....

eladkug (approved):
LGTM

PR #103, opened 2020-11-24
Add deployment design document

sigalsax:
### What does this PR do?

Adds deployment design document

MD: https://github.com/cyberark/conjur-api-python3/blob/add_deployment_design/design/deployment_design.md

### DIFF
--- a/design/deployment_design.md
+++ b/design/deployment_design.md
@@ -1,0 +1,102 @@
+#  **Dockerless CLI -** deployment
+
+This document is the design document for the deployment of the Python CLI. 
+
+Currently, our Python CLI can be deployed in the following ways:
+
+1. Install the CLI via the pip Python package manager like so: `pip3 install conjur-client`. **Python is required** on
+machine if installing the CLI in this way. If installed via pip, the user can then run the CLI like so `conjur-cli --help`
+
+2. Install and run the executable (**Python is** **not** required on the machine)
+
+   The CLI is delivered as a standalone executable, packed with the required runtime and dependencies (excluding a few
+   base linked libraries that are expected on the system). Therefore, there is no need for Python or other prerequisites
+   to be installed on machine. The executable can be fetched by navigating to our
+   [Python CLI release page](https://github.com/cyberark/conjur-api-python3/releases/tag/v0.0.5).
+   
+   If installed in this manner the CLI can be invoked by, `./python-cli-executable-location variable get ...`
+
+### Design
+
+#### Pip3 package
+
+We make CLI package available to the public by pushing to a public repository called PyPi. We will maintain pushing our
+the Python CLI to pip3 for both customer's machines have internet access and for our development work.
+
+Our Python API is also delivered this way and PyPi is the main installation method for it so this should be retained.
+The infrastructure is already in-place so this would be at no additional effort.
+
+#### Standalone executable
+
+To install the executable, for Linux and macOS machines we will provide tar.gz files. For Windows machine we will
+provide a zip file that will be available on the [release page](https://github.com/cyberark/conjur-api-python3/releases)
+of our Github repository.
+
+tar.gz is a standard archieve format for macOS and other Linux systems and has been used to deliver software in
+[another one of our open source projects](https://github.com/cyberark/secretless-broker/releases/tag/v1.7.1). tar.gz
+is a better solution than zip for Linux and macOS machines because zip's complement, unzip, does not come packaged in
+the common distributions (RHEL for example). This would require that users install unzip on their machines to use our
+software which should be avoided. We want to use tar.gz because it is usable without needing any additional software
+as unzip would demand. Each achived file will have its respective executable as detailed below.
+
+```bash
+conjur-cli-linux.tar.gz
+	conjur
+conjur-cli-darwin.tar.gz
+  conjur
+conjur-cli-windows.zip
+  conjur.exe
+```
+
+Once the user has installed the archive file, they will need to extract the contents by running
+`tar -xf conjur-cli-linux.tar.gz`
+
+Once they do this, they can run the newly extracted executable in one of the following ways:
+
+1. Run the executable from where the extraction took place like so: `./conjur --help` 
+2. Move the executable  (`conjur`) to their `/usr/local/bin` path so that it would look something like
+`/usr/local/bin/conjur`. An additional option would be having the user add it to their `PATH`. That way, the CLI
+executable will be available from anywhere on their machine without having to detail the full path. The exact location
+is left up to the user but we can provide a recommendation
+
+### Delivery
+
+At this time, the packing the CLI into an executable will be a manual process. We need to support RHEL, Windows, and
+macOS machines so we will manually spin up VMs (for RHEL and Windows) and manually pack the CLI:
+`pyinstaller --onefile pkg_bin/conjur`. On macOS we will pack locally.
+
+Once we have manually packed the executable and are ready for release, we will manually push the executables to the
+repo's release page.
+
+#### Upgrade from Ruby to Python CLI
+
+For all customers who have installed the Ruby CLI gem, we will request that the customer either remove it or rename it.
+They can do so doing the following:
+
+1. Run `which conjur` and they will get the path of the gem. For example `/Users/myuser/.rvm/gems/ruby-2.x.x/bin/conjur`
+2. Rename the gem found in that path to `/Users/myuser/.rvm/gems/ruby-2.x.x/bin/conjur.v6`
+
+### Testing
+
+A full test plan will be provided in a separate document in this repo. 
+
+### Helpful resources
+
+If in further phases we need to sign the files, see how to for macOS
+[here](https://github.com/pyinstaller/pyinstaller/wiki/Recipe-Win-Code-Signing) and here
+for [Windows](https://github.com/pyinstaller/pyinstaller/wiki/Recipe-Win-Code-Signing).
+
+### Open questions
+
+1. Do we need to sign both the zip and the executables? Not for phase one.
+
+### Delivery plan
+
+1. Implement changes outlined in the design (*3 days*)
+   1. Update executable and package name
+2. Map out the manual tests *(1 days)*
+3. Create documentation in `README.md` of the repository *(1 days)*
+4. Write draft document for TW for online help and open a docs ticket for TW *(2 days)*
+
+Total 7 days
+


design/deployment_design.md
```diff
@@ -0,0 +1,70 @@
+#  **Dockerless CLI -** deployment
+
+This document is the design document for the deployment of the Python CLI. 
+
+Currently, our Python CLI can be deployed in the following ways:
+
+1. Install the CLI via the pip Python package manager like so: `pip3 install conjur-client`. If installed in this way, Python **is** required on machine. If installed via pip, the user can then run the CLI like so `conjur-cli --help`
```
sgnn7, line 7: All lines over 90ish characters should be split

sgnn7, line 7:
```suggestion
1. Install the CLI via the pip Python package manager like so: `pip3 install conjur-client`. If installed in this way, **Python3 is required** on the machine if installing the CLI this way. If installed via pip, the user can then run the CLI like so `conjur-cli --help`
```

design/deployment_design.md
```diff
@@ -0,0 +1,70 @@
...
+
+2. Install and run the executable (Python **is** **not** required on the machine)
+
+   The CLI is delivered as a standalone executable, packed with the required runtime and dependencies. Therefore, there is no need for Python or other prerequisites to be installed on machine. The executable can be fetched by navigating to our [Python CLI release page](https://github.com/cyberark/conjur-api-python3/releases/tag/v0.0.5). If installed in this manner, `./python-cli-executable-location variable get ...`
```
sgnn7, line 11:
```suggestion
   The CLI is delivered as a standalone executable, packed with the required runtime and dependencies (excluding a few base linked libraries that are expected on the system). Therefore, there is no need for Python or other prerequisites to be installed on machine. The executable can be fetched by navigating to our [Python CLI release page](https://github.com/cyberark/conjur-api-python3/releases/tag/v0.0.5). If installed in this manner, the CLI can. be invoked with `./python-cli-executable-location variable get ...`
```

design/deployment_design.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Design
+
+#### Pip3 package
+
+We make CLI package available to the public by pushing to a public repository called PyPi. I propose that we maintain pushing our the Python CLI to pip3 for both customer's machines have have internet access and for our development work. The infrastructure is already in-place so this would be at no additional effort.
```
sgnn7, line 17:
```suggestion
We make CLI package available to the public by pushing to a public repository called PyPi. I propose that we maintain pushing our the Python CLI to pip3 for both customer's machines have have internet access and for our development work. Our Python API is also delivered this way and PyPi is the main installation method for it. The infrastructure is already in-place so this would be at no additional effort.
```

sgnn7 (commented):
@sigalsax Left a few comments

design/deployment_design.md
sigalsax: What types of base linked libraries? Like C level libraries?

InbalZilberman: @sigalsax the requirement was that we will have conjur cli named conjur can we have a different release mechanism to match this requirement? For example have folders in the release page
Mac
  -conjur
RHEL
  - conjur

This is very important requirement we need to meet

design/deployment_design.md
InbalZilberman: @sgnn7 What are the implication for having the Python API?
Can we not deliver it?

design/deployment_design.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+conjur.exe (windows)
+conjur (linux)
+conjur-darwin (macOS)
+```
+
+For Windows, executables have the `.exe` in their name so we will follow suit for the CLI Windows executable. 
+
+For Linux and macOS, we will need to differentiate them. For Linux, we can call the executable `conjur` and for macOS, `conjur-darwin`. That way, the common user-case (Linux) will have a cleaner experience.
```
InbalZilberman, line 31: The requirement is that anywhere you go the cli called conjur. Can we change the release structure to meet this requirement?

design/deployment_design.md
```diff
@@ -0,0 +1,70 @@
...
+conjur-darwin (macOS)
+```
+
+For Windows, executables have the `.exe` in their name so we will follow suit for the CLI Windows executable. 
+
+For Linux and macOS, we will need to differentiate them. For Linux, we can call the executable `conjur` and for macOS, `conjur-darwin`. That way, the common user-case (Linux) will have a cleaner experience. 
+
+A con of this approach would be that on the [release page](https://github.com/cyberark/conjur-api-python3/releases), at face value, the user will not be able to tell which executable is for linux machines. They will need to make an assumption because "linux" is not in the name as with macOS (-darwin) and Window (.exe) executables.
+
+Another option would be pushing zipped folders for each platform to the release page for our Github repository. Each zipped folder will have its respective executable as detailed below.
```
InbalZilberman, line 35: zips also need to be signed. I think folders are actually a nice option WDYT?
@Tovli @eladkug WDYT?

design/deployment_design.md
rafis3: I agree, they all need to be called conjur. In Windows, yes the exe is required in the end, but the UX remains the same. You can call `conjur` and it works.
If the naming duplication concerns you, I think that the folders approach is the right way.

design/deployment_design.md
```diff
@@ -0,0 +1,70 @@
...
+windows/conjur-cli.zip
+  conjur.exe
+```
+
+Once the user has installed the execuable, they can run the executable in one of the following ways:
+
+1. Run the executable like so: `./linux/conjur-cli/conjur --help` 
+2. Move the ***only*** executable (`conjur`) to their `/usr/local/bin` path so that it would look something like `/usr/local/bin/conjur`. That way, the CLI executable will be available from anywhere on their machine without having to detail the full path
+
+TODO: See if part of the installation process we can automatically save executable to the `/usr/local/bin` directory.
```
rafis3, line 51: There is more than one possible location for a binary. It depends on the user and how they work. I think the simplest would be letting the user decide where to unzip it and how to use it.

design/deployment_design.md
eladkug: I agree @InbalZilberman. In addition to execution file we can add specific README file on each folder (if something is relevant for specific OS.)

design/deployment_design.md
sigalsax: I still don't understand what split means. Just an 'enter' to move to a new line?

sigalsax: sounds a bit weird because if installing in this way appears at the beginning and end of the sentence. I changed it to:

```
**Python is required** on machine if installing the CLI in this way
```
is this clearer?

design/deployment_design.md
sigalsax: Sure I get that later in the doc. Here I am just stating what we **_currently_** have

design/deployment_design.md
sigalsax: Why do zips need to be signed by folders don't?
@InbalZilberman @eladkug Sounds good but don't think we push folders the each release. They need to be zipped. Or we can deliver the folders in a different way that is not through the release page of the repo

design/deployment_design.md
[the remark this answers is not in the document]
sigalsax: > This package basically combines both the Conjur Python SDK and the CLI, right?

Yes. 
Regarding splitting into the logic into two repos, how much of a development investment do you think it would it be?

jvanderhoof:
I really dislike the idea of having the Conjur CLI distributed as a zip file that needs to be unpacked and moved.  Personally, when I need to do this, it feels like I'm working with a one-off script, not a tool I'll be using regularly.  Have you looked at RVM as a cleaner example? To install:
```
curl -sSL https://get.rvm.io | bash
```
The result is RVM installed in a system folder where `rvm ...` can be called from anywhere in the file system. Users concerned with running random executables on their system can read the output to understand what the installer will be doing.

design/deployment_design.md
sgnn7: yup!

design/deployment_design.md
sgnn7: > What types of base linked libraries? Like C level libraries?

Yes.
```sh-session
root@d6f1f9e4afda:/# ldd conjur-py3-cli-linux 
        linux-vdso.so.1 (0x00007ffe365c7000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f4c62566000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f4c6254a000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f4c62358000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f4c62570000)
```
We can also use [Cython](https://cython.org/) if we need this list to be even smaller.

design/deployment_design.md
sgnn7: I was just noting here that the python client library is delivered to our users via PyPi too and it's the [primary way of using it](https://github.com/cyberark/conjur-api-python3#from-pypi). We must retain PyPi publishing if we want that to continue.

sgnn7:
I'd probably agree with @jvanderhoof that we also need some sort of installation script that you can pipe to bash regardless of our delivery method since that is what most people expect nowadays (even if it's not really the most appropriate app delivery method).

sigalsax:
> I really dislike the idea of having the Conjur CLI distributed as a zip file that needs to be unpacked and moved. Personally, when I need to do this, it feels like I'm working with a one-off script, not a tool I'll be using regularly. Have you looked at RVM as a cleaner example? To install:
> 
> ```
> curl -sSL https://get.rvm.io | bash
> ```
> 
> The result is RVM installed in a system folder where `rvm ...` can be called from anywhere in the file system. Users concerned with running random executables on their system can read the output to understand what the installer will be doing.

It is important to note here that we cannot use curl as an official method because we cannot assume customers will have internet access. Are there other methods that would not require internet access?

design/deployment_design.md
```diff
@@ -0,0 +1,70 @@
...
+1. Run `which conjur` and they will get the path of the gem. For example `/Users/myuser/.rvm/gems/ruby-2.x.x/bin/conjur`
+2. Rename the gem found in that path to `/Users/myuser/.rvm/gems/ruby-2.x.x/bin/conjur.v6`
+
+### Testing
+
+TBD
+
+### Open questions
+
+1. When installing the package via pip3, why is it called `conjur-client` instead of `conjur-cli`?
```
liavyona, line 70: Because it is not just a cli. It is a library. In python you can consume library by `import`. Which makes sense because I `import conjur_client` in my python program in order to use it (for example: https://github.com/conjurinc/aam-saas-conjur-enterprise/blob/main/scripts/conjur_sanity_check.py). But for every library I can define entry points (see `setup.py`) which will be mapped to `/usr/local/bin/` after installation. Thus, the library nameis conjur client because it provides a Conjur server client wrapper, while there is an additional module in the library which is a cli thus called `conjur-cli`

liavyona:
Another point:
You can install the package without pip (`python setup.py install .`)

design/deployment_design.md
```diff
@@ -0,0 +1,104 @@
...
+
+   The CLI is delivered as a standalone executable, packed with the required runtime and dependencies (excluding a few base linked libraries that are expected on the system). Therefore, there is no need for Python or other prerequisites to be installed on machine. The executable can be fetched by navigating to our [Python CLI release page](https://github.com/cyberark/conjur-api-python3/releases/tag/v0.0.5). If installed in this manner the CLI can be invoked by, `./python-cli-executable-location variable get ...`
+
+### Design
+
+#### Pip3 package
+
+We make CLI package available to the public by pushing to a public repository called PyPi. I propose that we maintain pushing our the Python CLI to pip3 for both customer's machines have have internet access and for our development work. 
+
+Our Python API is also delivered this way and PyPi is the main installation method for it so this should be retained. The infrastructure is already in-place so this would be at no additional effort.
```
Tovli, line 19: Python API is our SDK?

design/deployment_design.md
Tovli: What about adding link to the relevant release page? [see how dotnet CLI did it](https://github.com/dotnet/cli/releases)

design/deployment_design.md
sgnn7: @Tovli Yes, it's a wording issue. Our repos are called "conjur-api-<language>" but effectively it's an SDK for that language. In this case, when you see "Python API", what's really meant is "Python SDK" or "Python Client Library"

design/deployment_design.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Standalone executable
+
+To install the executable, we will provide tar.gz files for each platform on the [release page](https://github.com/cyberark/conjur-api-python3/releases) of our Github repository. tar.gz is a standard archieve format for macOS and other Linux systems and has been used to deliver software in [another one of our open source projects](https://github.com/cyberark/secretless-broker/releases/tag/v1.7.1).
+
+tar.gz is a better solution than zip because it's complement, `unzip`, does not come packaged in the common distributions (RHEL for example). This would require that users install `unzip` on their machines to use our software which should be avoided. We want to use tar.gz because it is usable without needing any additional software as unzip would demand. Each achieved file will have its respective executable as detailed below.
+
+```bash
+linux/conjur-cli.tar.gz
```
InbalZilberman, line 30: @sigalsax lets have no structure of folder in the tar
meaning once i run
tar -xf dawin-conjur-cli.tar.gz
i can simply run 
conjur --help
to see if it is working

InbalZilberman (commented):
Thank you!

design/deployment_design.md
sigalsax: @InbalZilberman FYI

design/deployment_design.md
Tovli: IMO - We can also direct users to add the CLI path to the `path` env var so they can run it from anywhere without moving it to a location that is already in this env var

sigalsax: I am 👌  with that

design/deployment_design.md
```diff
@@ -0,0 +1,79 @@
...
+
+   The CLI is delivered as a standalone executable, packed with the required runtime and dependencies (excluding a few base linked libraries that are expected on the system). Therefore, there is no need for Python or other prerequisites to be installed on machine. The executable can be fetched by navigating to our [Python CLI release page](https://github.com/cyberark/conjur-api-python3/releases/tag/v0.0.5). 
+   
+   If installed in this manner the CLI can be invoked by, `./python-cli-executable-location variable get ...`
+
+### Design
+
+#### Pip3 package
+
+We make CLI package available to the public by pushing to a public repository called PyPi. We will maintain pushing our the Python CLI to pip3 for both customer's machines have have internet access and for our development work.
```
eladkug, line 19: have - repeat twice

eladkug (dismissed):

Tovli (approved):
LGTM

PR #123, opened 2020-12-23
Bug fixes found from init manual tests

sigalsax:
### What does this PR do?
This PR fixes the bugs that were found from our manual tests

### DIFF
--- a/conjur/cli.py
+++ b/conjur/cli.py
@@ -38,6 +38,19 @@ class Cli():
   {}'''.format(*args)
 
     @staticmethod
+    def command_description(example, usage):
+        """
+        This method builds the header for the main screen.
+        """
+        return '''
+        
+Name:
+  {}
+
+Usage:
+  {}'''.format(example, usage)
+
+    @staticmethod
     def main_epilog():
         """
         This method builds the footer for the main help screen.
@@ -78,8 +91,8 @@ Copyright 2020 CyberArk Software Ltd. All rights reserved.
         test sources. Parses CLI args and invokes the appropriate client command.
         """
         formatter_class = lambda prog: argparse.RawTextHelpFormatter(prog,
-                                                                     max_help_position=60,
-                                                                     width=60)
+                                                                     max_help_position=100,
+                                                                     width=100)
         # pylint: disable=line-too-long
         parser = ArgparseWrapper(description=self.usage('conjur [global options] <command> <subcommand> [options] [args]'),
                                  epilog=self.main_epilog(),
@@ -90,10 +103,12 @@ Copyright 2020 CyberArk Software Ltd. All rights reserved.
         global_optional = parser.add_argument_group("Global options")
         resource_subparsers = parser.add_subparsers(dest='resource', title=self.title("Commands"))
 
+        init_example= 'init - Initialize Conjur configuration'
+        input_usage='conjur [global options] init [options] [args]'
         # pylint: disable=line-too-long
         init_subparser = resource_subparsers.add_parser('init',
                                                         help='Initialize the Conjur configuration',
-                                                        description=self.usage('conjur [global options] init [options] [args]'),
+                                                        description=self.command_description(init_example, input_usage),
                                                         epilog=self.command_epilog('conjur init -a my_org -u https://localhost\t'
                                                                                    'Initializes Conjur configuration and writes to file (.conjurrc)'),
                                                         usage=argparse.SUPPRESS,
@@ -107,7 +122,7 @@ Copyright 2020 CyberArk Software Ltd. All rights reserved.
                                   '(obtained from Conjur server unless provided by this option)')
         init_options.add_argument('-c', '--certificate',
                                   action='store', dest='certificate',
-                                  help='Provide Conjur SSL certificate file location' \
+                                  help='Provide Conjur SSL certificate file location ' \
                                   '(obtained from Conjur server unless provided by this option)')
         init_options.add_argument('--force',
                                   action='store_true',
@@ -183,17 +198,18 @@ Copyright 2020 CyberArk Software Ltd. All rights reserved.
 
         resource, args = Cli._parse_args(parser)
 
-        # TODO Add tests for exception handling logic
         # pylint: disable=broad-except
         try:
             Cli.run_client_action(resource, args)
+        except KeyboardInterrupt:
+            sys.exit(0)
         except FileNotFoundError as not_found_error:
             logging.debug(traceback.format_exc())
             sys.stdout.write(f"Error: No such file or directory: '{not_found_error.filename}'")
             sys.exit(1)
         except Exception as error:
             logging.debug(traceback.format_exc())
-            sys.stdout.write(str(error))
+            sys.stdout.write(f"{str(error)}\n")
             sys.exit(1)
         else:
             # Explicit exit (required for tests)
@@ -214,7 +230,7 @@ Copyright 2020 CyberArk Software Ltd. All rights reserved.
 
             # A successful exit is required to prevent the initialization of
             # the Client because the init command does not require the Client
-            sys.exit(0)
+            return
 
         client = Client(ssl_verify=args.ssl_verify, debug=args.debug)
 
--- a/conjur/constants.py
+++ b/conjur/constants.py
@@ -11,4 +11,4 @@ import os
 
 DEFAULT_CONFIG_FILE = os.path.expanduser(os.path.join('~', '.conjurrc'))
 DEFAULT_NETRC_FILE = os.path.expanduser(os.path.join('~', '.netrc'))
-CERTIFICATE_FILENAME = "conjur-server.pem"
+DEFAULT_CERTIFICATE_FILE=os.path.expanduser(os.path.join('~', "conjur-server.pem"))
--- a/conjur/init/init_controller.py
+++ b/conjur/init/init_controller.py
@@ -9,7 +9,6 @@ required to successfully configure the conjurrc
 
 # Builtins
 import logging
-import os
 import sys
 
 # Third party
@@ -45,8 +44,8 @@ class InitController:
         self.write_conjurrc()
 
         sys.stdout.write("Configuration initialized successfully!\n")
-        sys.stdout.write("To begin using the CLI, log in to the Conjur server by" \
-                         "running `conjur authn login`")
+        sys.stdout.write("To begin using the CLI, log in to the Conjur server by " \
+                         "running `conjur login`\n")
 
     def get_server_certificate(self):
         """
@@ -54,15 +53,27 @@ class InitController:
         """
         # pylint: disable=line-too-long
         if self.conjurrc_data.appliance_url is None:
-            self.conjurrc_data.appliance_url = input("Enter the URL of your Conjur server: ")
+            self.conjurrc_data.appliance_url = input("Enter the URL of your Conjur server: ").strip()
+            if self.conjurrc_data.appliance_url == '':
+                # pylint: disable=raise-missing-from
+                raise RuntimeError("Error: URL is required")
+
 
         url = urlparse(self.conjurrc_data.appliance_url)
-        if url.scheme == "http":
-            return None
+
+        # TODO: Factor out the following URL validation to ConjurrcData class
+        # and add integration tests
+        # At this time, providing ports is not supported and
+        # all urls must start with HTTPS.
+        if url.port is not None or url.scheme != 'https':
+            raise RuntimeError(f"Error: undefined behavior. Reason: The Conjur url format provided "
+                   f"'{self.conjurrc_data.appliance_url}' is not supported. "
+                   "Consider adding HTTPS as the prefix and remove the port if provided")
 
         if self.conjurrc_data.cert_file is not None:
             # Return None because we do not need to fetch the certificate
             return None
+
         # pylint: disable=logging-fstring-interpolation
         logging.debug(f"Initiating a TLS connection with '{self.conjurrc_data.appliance_url}'")
         fingerprint, fetched_certificate = self.init_logic.get_certificate(url.hostname, url.port)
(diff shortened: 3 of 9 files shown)


conjur/cli.py
```diff
@@ -93,7 +106,7 @@ def run(self, *args):
         # pylint: disable=line-too-long
         init_subparser = resource_subparsers.add_parser('init',
                                                         help='Initialize the Conjur configuration',
-                                                        description=self.usage('conjur [global options] init [options] [args]'),
+                                                        description=self.command_description('init - Initialize Conjur configuration', 'conjur [global options] init [options] [args]'),
```
abrahamko, line 109: do not inline in inside this call, move it to a local variable.

conjur/cli.py
```diff
@@ -183,10 +196,11 @@ def run(self, *args):
 
         resource, args = Cli._parse_args(parser)
 
-        # TODO Add tests for exception handling logic
         # pylint: disable=broad-except
         try:
             Cli.run_client_action(resource, args)
+        except KeyboardInterrupt:
+            sys.exit(0)
```
abrahamko, line 203: suggestion: add debug log?

conjur/cli.py
sigalsax: the string inside?

conjur/cli.py
sigalsax: Not critical. I think the user seeing their `^C` is enough

conjur/constants.py
```diff
@@ -12,3 +12,4 @@
 DEFAULT_CONFIG_FILE = os.path.expanduser(os.path.join('~', '.conjurrc'))
 DEFAULT_NETRC_FILE = os.path.expanduser(os.path.join('~', '.netrc'))
 CERTIFICATE_FILENAME = "conjur-server.pem"
+FULL_CERTIFICATE_PATH=os.path.join(os.path.dirname(DEFAULT_CONFIG_FILE), CERTIFICATE_FILENAME)
```
abrahamko, line 15:
```suggestion
FULL_CERTIFICATE_PATH=os.path.expanduser(os.path.join('~', CERTIFICATE_FILENAME))
```

conjur/constants.py
```diff
@@ -12,3 +12,4 @@
 DEFAULT_CONFIG_FILE = os.path.expanduser(os.path.join('~', '.conjurrc'))
 DEFAULT_NETRC_FILE = os.path.expanduser(os.path.join('~', '.netrc'))
 CERTIFICATE_FILENAME = "conjur-server.pem"
```
abrahamko, line 14: I think this const is unnecessary.

sigalsax: okay I will remove

conjur/init/init_controller.py
```diff
@@ -45,24 +45,56 @@ def load(self):
...
+                raise RuntimeError("Error: URL is required\n")
+
 
         url = urlparse(self.conjurrc_data.appliance_url)
+        # TODO: Factor out the following URL validation to ConjurrcData class
+        # pylint: disable=pointless-string-statement
+
+        # This condition catches if a user provides input <hostname>
+        # without HTTP/HTTPS prefix. If so, we default the behavior
+        # to HTTPS because this is what we currently support
```
abrahamko, line 69: This is not the correct place to do these validations.
Create a method `get_conjur_url`, and then validate the input there.

conjur/cli.py
sigalsax: moved string to own var

conjur/constants.py
sigalsax: updating

conjur/init/init_controller.py
```diff
@@ -45,24 +45,56 @@ def load(self):
...
+                               "Consider adding HTTPS as the prefix\n")
+
         if self.conjurrc_data.cert_file is not None:
             # Return None because we do not need to fetch the certificate
             return None
+
+        # To follow suit with the Ruby CLI, we don't write the url and the port
+        # in the conjurrc. If the port is added to the conjurrc following requests
+        # will fail
+        self.conjurrc_data.appliance_url = url.scheme + '://' + url.hostname
```
abrahamko, line 96: You can't ignore the port, removing the port means you refer to different location than the one specified.
If you only support portless http/https, validate it and fail, so you won't need to write this kind of code.

conjur/init/init_controller.py
```diff
@@ -71,9 +103,10 @@ def get_server_certificate(self):
         sys.stdout.write("\nTo verify this certificate, it is recommended to run the following " \
                          "command on the Conjur server:\n" \
                          "openssl x509 -fingerprint -noout -in ~conjur/etc/ssl/conjur.pem\n\n")
-        trust_certificate = input("Trust this certificate? [no]: ")
+        trust_certificate = input("Trust this certificate? [no]: ").strip()
```
abrahamko, line 106: can `input` return None?

conjur/init/init_controller.py
```diff
@@ -87,23 +120,22 @@ def get_account_info(self, conjurrc_data):
...
+                    raise RuntimeError("Error: account is required\n")
 
     def write_certificate(self, fetched_certificate):
         """
         Method to write the certificate fetched from the Conjur endpoint on the user's machine
         """
-        url = urlparse(self.conjurrc_data.appliance_url)
-        if self.conjurrc_data.cert_file is None and url.scheme == "https":
+        # pylint: disable=line-too-long
+        if self.conjurrc_data.cert_file is None and self.conjurrc_data.appliance_url.startswith("https"):
```
abrahamko, line 136: this change is unnecessary, the previous code was better.
Don't try to parse the url yourself, use the dedicated lib for that.

conjur/init/init_controller.py
```diff
@@ -113,7 +145,7 @@ def write_certificate(self, fetched_certificate):
                                                           self.conjurrc_data.cert_file,
                                                           True)
 
-            sys.stdout.write(f"Certificate written to {conjur.constants.CERTIFICATE_FILENAME}\n\n")
+            sys.stdout.write(f"Certificate written to {conjur.constants.FULL_CERTIFICATE_PATH}\n\n")
```
abrahamko, line 148: Good change!

conjur/init/init_controller.py
```diff
@@ -132,6 +164,6 @@ def write_conjurrc(self):
     @staticmethod
     def __ensure_overwrite_file(config_file):
         force_overwrite = input(f"File {config_file} exists. " \
-                                f"Overwrite [yes]: ")
+                                f"Overwrite? [yes]: ").strip()
         if force_overwrite != '' and force_overwrite.lower() != 'yes':
-            raise Exception(f"Not overwriting {config_file}")
+            raise Exception(f"Not overwriting {config_file}\n")
```
abrahamko, line 169: error messages should never contain newlines.
If you need newlines in the output, add them where you write the output.

conjur/init/init_logic.py
```diff
@@ -42,7 +42,7 @@ def get_certificate(self, hostname, port):
             logging.debug("Successfully fetched certificate")
         except Exception as error:
             raise Exception(f"Unable to retrieve certificate from {hostname}:{port}. " \
-                            f"Reason: {str(error)}") from error
+                            f"Reason: {str(error)}\n") from error
```
abrahamko, line 45: same as above.

conjur/init/init_controller.py
sigalsax: The Ruby CLI doesn't add the port to the conjurrc. I use that port to fetch the certificate 
Should we just not support a port?

conjur/init/init_controller.py
sigalsax: No it will error if it is empty. Returning `You decided not to trust the certificate`

test/test_ssl_service.py
```diff
@@ -5,6 +5,11 @@
 
 from conjur.ssl_service import SSLService
 
+"""
+*************** DEVELOPER NOTE ***************
```
abrahamko, line 9: Not relevant for this PR.
Also, I meant to add message to the assert itself, so when it fails it will be obvious why.

conjur/init/init_controller.py
sigalsax: reverting

abrahamko (changes requested):
The error handling should improve to get approval.

conjur/init/init_controller.py
abrahamko: yes. Validate no port is supplied, fail it if it does, and do not build the URL yourself.

test/test_ssl_service.py
sigalsax: as in add the comment above the assert like so:
```
        # These certificate will expire in 2022 which means this test will fail.
        # If this is the case, update the expected output
        self.assertEquals(fingerprint, GITHUB_FINGERPRINT)
        self.assertEquals(readable_certificate, GITHUB_CERT)
```

conjur/init/init_controller.py
```diff
@@ -71,9 +103,10 @@ def get_server_certificate(self):
         sys.stdout.write("\nTo verify this certificate, it is recommended to run the following " \
                          "command on the Conjur server:\n" \
                          "openssl x509 -fingerprint -noout -in ~conjur/etc/ssl/conjur.pem\n\n")
-        trust_certificate = input("Trust this certificate? [no]: ")
+        trust_certificate = input("Trust this certificate? [no]: ").strip()
         if trust_certificate.lower() != 'yes':
-            raise ValueError("You decided not to trust the certificate")
+            logging.debug("Invalid input. Acceptable input includes: 'YES', 'yes' ")
```
abrahamko, line 108: `no` is not invalid input.
I think this log is redundant anyway.

conjur/init/init_controller.py
sigalsax: What are the consequences of such a decision?

conjur/init/init_controller.py
sigalsax: added a todo

conjur/init/init_controller.py
sigalsax: removing

conjur/init/init_controller.py
```diff
@@ -45,24 +45,43 @@ def load(self):
...
 
     def get_server_certificate(self):
         """
         Method to get the certificate from the Conjur endpoint detailed by the user
         """
         # pylint: disable=line-too-long
         if self.conjurrc_data.appliance_url is None:
-            self.conjurrc_data.appliance_url = input("Enter the URL of your Conjur server: ")
+            self.conjurrc_data.appliance_url = input("Enter the URL of your Conjur server: ").strip()
+            if self.conjurrc_data.appliance_url is None or self.conjurrc_data.appliance_url == '':
```
abrahamko, line 58: It cannot be None here, it would throw when you call `strip()` in the previous line.
So You can remove it:
```suggestion
            if self.conjurrc_data.appliance_url == '':
```

conjur/init/init_controller.py
```diff
@@ -45,24 +45,43 @@ def load(self):
...
+                # pylint: disable=raise-missing-from
+                raise RuntimeError("Error: URL is required")
+
 
         url = urlparse(self.conjurrc_data.appliance_url)
+        # TODO: Factor out the following URL validation to ConjurrcData class
+        # At this time, providing ports is not supported.
+        # All HTTPS activity will be through 443
+        if url.netloc.find(":443") != -1 or (url.scheme != '' and url.path != ''):
```
abrahamko, line 67: Why not use `.port`?
```suggestion
        if url.port == '' or (url.scheme != '' and url.path != ''):
```

conjur/init/init_controller.py
```diff
@@ -45,24 +45,43 @@ def load(self):
...
+            raise RuntimeError(f"Error: undefined behavior. Reason: The Conjur url format provided "
+                   f"'{self.conjurrc_data.appliance_url}' is not supported. "
+                   "Consider adding HTTPS as the prefix or remove the port if provided")
+
+        if url.scheme == '':
+            parsed_url = urllib.parse.urlparse(self.conjurrc_data.appliance_url)
+            replaced_url = parsed_url._replace(scheme="https", netloc=url.path, path='')
```
abrahamko, line 74: Do not build the url yourself.
you can use the urllib for that.

conjur/init/init_controller.py
sigalsax: because https://something will also have a port = 443. Only when you attach that 443 to the url the netloc will have 443 which is why I have this condition

conjur/init/init_controller.py
```diff
@@ -45,24 +45,43 @@ def load(self):
...
+            url = replaced_url
+            self.conjurrc_data.appliance_url=replaced_url.geturl()
```
abrahamko, line 76: these 4 lines look very messy.
Concise and write exactly what you need.
I think you meant to add `https` if you don't have one.
urlparse might surprise you when there's no scheme.
You will have to test it very carefully, as this code can easily break.

conjur/cli.py
abrahamko: Not exactly what I meant, but it's good enough.

conjur/constants.py
```diff
@@ -11,4 +11,4 @@
 
 DEFAULT_CONFIG_FILE = os.path.expanduser(os.path.join('~', '.conjurrc'))
 DEFAULT_NETRC_FILE = os.path.expanduser(os.path.join('~', '.netrc'))
-CERTIFICATE_FILENAME = "conjur-server.pem"
+FULL_CERTIFICATE_PATH=os.path.expanduser(os.path.join('~', "conjur-server.pem"))
```
abrahamko, line 14: Why did you name it differently than the others?
It is a default path, and it's a file, so DEFAULT_CERTIFICATE_FILE seems correct to me.

abrahamko (dismissed):

abrahamko (approved):

PR #127, opened 2020-12-28
Add login/logout command logic

sigalsax:
### What does this PR do?
Adds login/logout functionality to the CLI

TODO
- [ ] Add UT/Integration tests
- [ ] Update changelog

#### Change log
- [ ] This PR does not include user-facing changes and doesn't require a CHANGELOG update

#### Test coverage

#### Documentation


conjur/cli.py
```diff
@@ -19,6 +19,7 @@
 # Internals
 from conjur.argparse_wrapper import ArgparseWrapper
 from conjur.client import Client
+from conjur.login import NetrcData, LoginLogic, LoginController, CredentialsFromFile
```
sigalsax, line 22: TODO: Might need to change the name Login<xyz> because this logic ideally covers both Login **and** Logout

conjur/logout/logout_controller.py
```diff
@@ -0,0 +1,45 @@
...
+from conjur.utils import Utils
+
+# pylint: disable=too-few-public-methods
+class LogoutController:
+    """
+    LogoutController
+
+    This class represents the Presentation Layer for the LOGOUT command
+    """
+    logout_logic = None
```
mbenita-Cyberark, line 24: why static?

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        with open(DEFAULT_NETRC_FILE, 'w') as netrc_file:
+            # pylint: disable=line-too-long
+            netrc_file.write(f"machine {credential_data.machine}\nlogin {credential_data.login}\npassword {credential_data.password}\n")
+            netrc_file.close()
+
+        # TODO validate this isn't a problem on Windows
+        # TODO validate _netrc on Windows
+        # Ensures that the netrc file is only available to owner
+        os.chmod(DEFAULT_NETRC_FILE, stat.S_IRWXU)
```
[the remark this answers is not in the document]
sigalsax: If I am understanding correctly, moving `api_key`, `login_id` to a constructor of the class and the return of load will still be a dictionary?
```
def __init__(api_key, login_id):
  self.api_key=api_key
...
def load()
  loaded_netrc[self.api_key]= api_key
```
?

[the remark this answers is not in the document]
sigalsax: >  instead of "credential_data" we will have a strong typed data that we get from the user in init

Not sure what you mean here because load data doesn't get filled from the user, but from the actual file (netrc)

conjur/login/credential_data.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __init__(self, machine=None, login=None, password=None):
+        self.machine = machine
+        self.login = login
+
+        # Purposefully not named the same for clarity in code. .netrc file format standard
+        # uses password but this value is actually the user/host api key. This convention should
+        # be kept. See https://www.labkey.org/Documentation/wiki-page.view?name=netrc especially
+        # the 'Use API Keys' section
+        self.password = password
```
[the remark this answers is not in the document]
sigalsax: It could be None if the user didn't provide a password but the user wouldnt get past  `get_api_key()` which will fail on login and the netc won't be created

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
...
+from conjur.utils import Utils
+import conjur.constants
+
+class LoginController:
+    """
+    LoginController
+
+    This class represents the Presentation Layer for the LOGIN command
+    """
+    password = None
```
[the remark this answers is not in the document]
sigalsax: I was debating where to call the credential_data property api_key or password for this exact reason. The password referenced here is the password or api key the user or host inputs. In the actual credentials_data object regardless the value will be the api_key fetched from login.

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
...
+                raise RuntimeError("Error: Login name is required")
+
+    def get_password(self):
+        """
+        Method to fetch the password from the user attempting to login
+        """
+        if self.password is None:
+            # pylint: disable=line-too-long
+            return getpass.getpass(prompt=f"Please enter {self.credential_data.login}'s password (it will not be echoed): ")
+        return self.password
```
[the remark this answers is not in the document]
sigalsax: no because even if it is None, we will still attempt to make a login request. We could block empty input but didn't think it was necessary since it is only a single login request that is made

conjur/logout/logout_controller.py
sigalsax: removing 😅

conjur/cli.py
```diff
@@ -231,6 +292,34 @@ def run_client_action(resource, args):
...
+        elif resource == 'login':
+            credential_data = CredentialData(None, args.name, None)
+            credentials = CredentialsFromFile()
+            login_logic = LoginLogic(credentials)
+            login_controller = LoginController(ssl_verify=args.ssl_verify,
+                                               password=args.password,
+                                               credential_data=credential_data,
+                                               login_logic=login_logic)
+            login_controller.load()
+            return
```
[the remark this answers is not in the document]
abrahamko: I agree

conjur/cli.py
```diff
@@ -231,6 +292,34 @@ def run_client_action(resource, args):
...
+        elif resource == 'logout':
+            credentials = CredentialsFromFile()
+            logout_logic = LogoutLogic(credentials)
+
+            logout_controller = LogoutController(ssl_verify=None,
+                                                 logout_logic=logout_logic)
```
abrahamko, line 310: I think this i redundant, just `logout_logic is enough.
```suggestion
                                                 logout_logic)
```

conjur/cli.py
```diff
@@ -231,6 +292,34 @@ def run_client_action(resource, args):
...
+            logout_controller.remove_credentials()
+            return
+        # pylint: disable=line-too-long
+        if not os.path.exists(DEFAULT_NETRC_FILE) or (os.path.exists(DEFAULT_NETRC_FILE) and os.path.getsize(DEFAULT_NETRC_FILE) == 0):
```
abrahamko, line 314: I don't understand.
You do login logic for every command?

abrahamko, line 314: No need to check for file non-existence and then for file existence. you can remove the second check:
```suggestion
        if not os.path.exists(DEFAULT_NETRC_FILE) or os.path.getsize(DEFAULT_NETRC_FILE) == 0:
```

conjur/cli.py
```diff
@@ -231,6 +292,34 @@ def run_client_action(resource, args):
...
+            credential_data = CredentialData(None, None, None)
```
abrahamko, line 315: Put `None` as default values on the parameters, so you won't have to explicitly pass them here.

conjur/cli.py
```diff
@@ -272,7 +361,7 @@ def _parse_args(parser):
             sys.exit(0)
 
         # Check whether we are running a command with required additional arguments/options
-        if args.resource not in ['list', 'whoami', 'init']:
+        if args.resource not in ['list', 'whoami', 'init', 'login', 'logout']:
```
abrahamko, line 364: `init` and `logout` does not get here, so no need to put them.

conjur/client.py
```diff
@@ -104,10 +102,17 @@ def __init__(self,
                             **loaded_config)
             self._api.login(login_id, password)
         else:
-            logging.info("Using API key with netrc credentials...")
+            try:
+                loaded_netrc = CredentialsFromFile.load()
+
+            except Exception:
+                # pylint: disable=line-too-long, raise-missing-from
+                raise RuntimeError("Unable to authenticate with Conjur. Please log in and try again")
```
abrahamko, line 110: add `from ex`.

conjur/client.py
```diff
@@ -81,10 +80,9 @@ def __init__(self,
                         on_disk_config[field_name] = field_value
                 loaded_config = on_disk_config
 
+            # TODO add error handling for when conjurrc field doesn't exist
             except Exception as exc:
-                raise ConfigException(f"Error: {exc} not found in " \
-                                      "the loaded conjurrc\n") from Exception
-
+                raise ConfigException(exc) from Exception
```
abrahamko, line 85: Isn't it a bug?
Shouldn't it be like this?
```suggestion
                raise ConfigException(exc) from exc
```

abrahamko: For this you need code coverage tests that ensure you write interpretable code that behave as you intended.

conjur/constants.py
```diff
@@ -8,7 +8,17 @@
...
+    NETRC="_netrc"
+else:
+    NETRC=".netrc"
 
 DEFAULT_CONFIG_FILE = os.path.expanduser(os.path.join('~', '.conjurrc'))
-DEFAULT_NETRC_FILE = os.path.expanduser(os.path.join('~', '.netrc'))
-DEFAULT_CERTIFICATE_FILE=os.path.expanduser(os.path.join('~', "conjur-server.pem"))
+DEFAULT_NETRC_FILE = os.path.expanduser(os.path.join('~', NETRC))
+DEFAULT_CERTIFICATE_FILE = os.path.expanduser(os.path.join('~', "conjur-server.pem"))
+NETRC_HOST_PATH = "/authn"
```
abrahamko, line 24: How's NETRC related to this?

conjur/constants.py
```diff
@@ -8,7 +8,17 @@
...
 import os
+import platform
+
+# The OS the CLI is run on is determined by the following:
+# See https://stackoverflow.com/questions/1854/python-what-os-am-i-running-on
+# pylint: disable=no-member
+if os.name != "posix" and platform.system() == "Windows":
+    NETRC="_netrc"
+else:
+    NETRC=".netrc"
```
abrahamko, line 19: Why only NETRC has this logic?
Also, better create a prefix const with this logic, and append the file name to it.
```suggestion
if os.name != "posix" and platform.system() == "Windows":
    INTERNAL_FILE_PREFIX="_"
else:
    INTERNAL_FILE_PREFIX="."
```

And then you will have:
```python
DEFAULT_NETRC_FILE = os.path.expanduser(os.path.join('~', INTERNAL_FILE_PREFIX + "netrc"))
```

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @classmethod
+    def remove(cls):
+        """
+        Method that removes netrc data from the file.
+        Triggered during a logout
+        """
+        if os.path.exists(DEFAULT_NETRC_FILE) and os.path.getsize(DEFAULT_NETRC_FILE) != 0:
+            open(DEFAULT_NETRC_FILE, "w").close()
```
abrahamko, line 46: Why not deleting the file?

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+        else:
+            raise Exception("Please log in")
```
abrahamko, line 48: ???
This is not in this class responsibility to handle login/logout.
This method should remove the credentials form the file.
It's in the caller responsibility to handle failures.

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+# Internals
+from conjur.constants import DEFAULT_NETRC_FILE
+
+class CredentialsFromFile:
+    """
+    CredentialsFromFile
+
+    This class holds logic when credentials are kept in the netrc
+    """
+    @classmethod
```
abrahamko, line 25: In OOP, you can use object to handle this, instead of static class methods.
You could pass the credentials file path as ctor parameter, instead of referring to the const every time.
Its default value could be the const, so you won't have to supply it from outside.

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+    def save(cls, credential_data):
+        """
+        Method that writes user data to a netrc file
+        and updates permissions on the file
+        """
+        with open(DEFAULT_NETRC_FILE, 'w') as netrc_file:
+            # pylint: disable=line-too-long
+            netrc_file.write(f"machine {credential_data.machine}\nlogin {credential_data.login}\npassword {credential_data.password}\n")
```
abrahamko, line 33: Write multiline string using triple quotes

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+            netrc_file.close()
```
abrahamko, line 34: Do you need to call `close()`? doesn't the scope handles it for you?

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+        Method that removes netrc data from the file.
+        Triggered during a logout
+        """
+        if os.path.exists(DEFAULT_NETRC_FILE) and os.path.getsize(DEFAULT_NETRC_FILE) != 0:
+            open(DEFAULT_NETRC_FILE, "w").close()
+        else:
+            raise Exception("Please log in")
+
+    @classmethod
+    def load(cls):
```
abrahamko, line 51: Very confusing name.
You use `load` to start execution of flow in other classes.
Here you overload its meaning, not a good desicion.

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        Method that loads the netrc data.
+        Triggered before each CLI action
+        """
+        # pylint: disable=logging-fstring-interpolation
+        logging.info(f"Using netrc as credential store {DEFAULT_NETRC_FILE}")
```
abrahamko, line 57: since you don't use credentials store, this log line is more misleading than informative.
Better:
```suggestion
        logging.info(f"Using {DEFAULT_NETRC_FILE} file to save credential")
```

conjur/init/conjurrc_data.py
```diff
@@ -16,6 +29,17 @@ def __init__(self, appliance_url = None, account = None, cert_file = None):
...
+    @classmethod
+    def load_conjurrc(cls):
+        """
+        Method that loads the conjurrc into an object
+        """
+        if os.path.exists(conjur.constants.DEFAULT_CONFIG_FILE):
+            with open(conjur.constants.DEFAULT_CONFIG_FILE, 'r') as conjurrc:
+                return load(conjurrc, Loader=Loader)
+        else:
+            raise FileNotFoundError
```
abrahamko, line 41: Let the file opening throw this failure.

conjur/init/conjurrc_data.py
```diff
@@ -6,6 +6,19 @@
 This module represents an object that holds conjurrc data
 """
 # pylint: disable=too-few-public-methods
+
+# Third party
+import os
+
+from yaml import load
```
abrahamko, line 13: alias `load` to `yaml_load`

conjur/init/conjurrc_data.py
```diff
@@ -16,6 +29,17 @@ def __init__(self, appliance_url = None, account = None, cert_file = None):
...
         self.plugins = []
 
+    @classmethod
+    def load_conjurrc(cls):
+        """
+        Method that loads the conjurrc into an object
+        """
+        if os.path.exists(conjur.constants.DEFAULT_CONFIG_FILE):
+            with open(conjur.constants.DEFAULT_CONFIG_FILE, 'r') as conjurrc:
+                return load(conjurrc, Loader=Loader)
```
abrahamko, line 39: Loader=Loader is redundant.
```suggestion
                return load(conjurrc, Loader)
```

conjur/init/conjurrc_data.py
```diff
@@ -16,6 +29,17 @@ def __init__(self, appliance_url = None, account = None, cert_file = None):
         self.cert_file = cert_file
         self.plugins = []
 
+    @classmethod
+    def load_conjurrc(cls):
```
abrahamko, line 33: call this method `load_from_file`, and get the file path as parameter.
You can default it to `DEFAULT_CONFIG_FILE`, so you won't need to change the caller.

conjur/init/init_controller.py
```diff
@@ -98,7 +98,7 @@ def get_account_info(self, conjurrc_data):
             # pylint: disable=broad-except,logging-fstring-interpolation
             except Exception as error:
                 # pylint: disable=line-too-long,logging-fstring-interpolation
-                logging.debug(f"Unable to fetch the account from the Conjur server. Reason: {error}")
+                logging.warning(f"Unable to fetch the account from the Conjur server. Reason: {error}")
```
abrahamko, line 101:
```suggestion
                logging.warning(f"Failed to fetch the account from the Conjur server. Reason: {error}")
```

conjur/login/credential_data.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+
+"""
+CredentialData module
+
+This module represents an object that holds netrc data
+"""
+
+# pylint: disable=too-few-public-methods
+class CredentialData:
```
abrahamko, line 10: Credential**s** is the correct term, I think.
Make sure to reflect this change to all the code.
```suggestion
class CredentialsData:
```

conjur/login/credential_data.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    def __init__(self, machine=None, login=None, password=None):
+        self.machine = machine
+        self.login = login
+
+        # Purposefully not named the same for clarity in code. .netrc file format standard
+        # uses password but this value is actually the user/host api key. This convention should
+        # be kept. See https://www.labkey.org/Documentation/wiki-page.view?name=netrc especially
+        # the 'Use API Keys' section
```
abrahamko, line 22: this clarity isn't clear to me now... 😉 
Does it store password or API key?

conjur/login/credential_data.py
```diff
@@ -0,0 +1,29 @@
...
+        self.password = password
+
+    # pylint: disable=line-too-long
+    def __repr__(self):
+        return "{'machine': %r, 'login': %r, 'password': %r}" % (self.machine,
+                                                                 self.login,
+                                                                 self.password)
```
abrahamko, line 29: You need to obfuscate the password when you print it.

abrahamko: like print only first 2 letters and replace the rest with ellipsis:
```suggestion
                                                                 self.password[:2] + "...")
```
So if the password is `P@ssw0rd`, you will see: `P@...` in the print.

conjur/login/credential_data.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+
+"""
+CredentialData module
```
abrahamko, line 4: Doesn't credentialsData used outside of `login`?
If so take it out of this folder, since it doesn't belong here.

conjur/init/conjurrc_data.py
```diff
@@ -6,6 +6,19 @@
 This module represents an object that holds conjurrc data
 """
 # pylint: disable=too-few-public-methods
+
+# Third party
+import os
+
+from yaml import load
+try:
+    from yaml import CLoader as Loader
```
abrahamko, line 15:
```suggestion
    from yaml import CLoader as YamlLoader
```

conjur/init/conjurrc_data.py
```diff
@@ -6,6 +6,19 @@
...
+except ImportError:
+    from yaml import Loader
```
abrahamko, line 17:
```suggestion
    from yaml import YamlLoader
```

conjur/config.py
```diff
@@ -8,7 +8,6 @@
 """
 
 import logging
-import netrc
 
 from yaml import load, dump
```
abrahamko, line 12: Do you still use these?

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+        loaded_netrc = {}
+
+        netrc_obj = netrc.netrc(DEFAULT_NETRC_FILE)
```
abrahamko, line 60: Who owns this `netrc` object?
Isn't it part of this repo?
Consider doing one of the following:
1. merge your work with this object
2. delete this object and use yours
3. ignore it (that is, leave it as is) and write all the relevant logic here without referring to it.

This mix doesn't look good and degrades maintainability.

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,69 @@
...
+        # Get the host url from the netrc object
+        netrc_host_url = list(netrc_obj.hosts)[0]
+        netrc_auth = netrc_obj.authenticators(netrc_host_url)
+        login_id, _, api_key = netrc_auth
+
+        loaded_netrc['api_key'] = api_key
+        loaded_netrc['login_id'] = login_id
+
+        return loaded_netrc
```
abrahamko, line 69: Why returning hash and not `CredentialsData`?
It doesn't make sense.
If you need more data object types like `CredentialsData`, create them.

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
+# -*- coding: utf-8 -*-
+
+"""
+LoginController module
+
+This module is the controller that facilitates all login actions
+required to successfully configure the netrc
```
abrahamko, line 7: As I told you, you should remove all reference to `netrc` as a concept.
Use `Credentials` instead.

conjur/login/login_controller.py
abrahamko, line 73: It is redundant to return the object field.
It's already part of the object, no need to return it.

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
...
+        self.credential_data = credential_data
+        self.login_logic = login_logic
+
+    def load(self):
+        """
+        Method that facilitates all method calls in this class
+        """
+        self.get_username()
+        self.password = self.get_password()
+        loaded_conjurrc = self.load_conjurrc_data()
```
abrahamko, line 45: No need to call it `loaded`.
`conjurrc` is enough. (which is somewhat a weird name, as it says nothing about its content, but whatever...)

conjur/init/conjurrc_data.py
abrahamko, line 39: It should return `conjurrc_data` object.
It's really weird that it doesn't.

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    password = None
+    credential_data = None
+    login_logic = None
+    ssl_verify = None
+
+    def __init__(self, ssl_verify, password, credential_data, login_logic):
+        self.ssl_verify = ssl_verify
+        if self.ssl_verify is False:
+            Utils.insecure_message()
```
abrahamko, line 33: Method should be named as an action describing what it does, not as a noun.
Replace with similar to:
```suggestion
            Utils.print_insecure_message()
```

abrahamko: actually, it would be better to keep all the printing to the user inside the `controller` (remember?)
So you need something like:
```suggestion
            print(Utils.get_insecure_message())
```

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        try:
+            fetched_api_key = self.login_logic.get_api_key(ssl_verify,
+                                                           self.credential_data,
+                                                           self.password,
+                                                           loaded_conjurrc)
+            self.credential_data.password = fetched_api_key
+        except Exception as error:
+            raise RuntimeError("Failed to login. Unable to authenticate with Conjur. " \
+                f"Reason: {error}. Please check your credentials and log in again") from error
```
abrahamko, line 98: Since you don't know the failure reason, you can't suggest to `check your credentials and log in again`.
The failure might be due to bad gateway, or other reasons.
Better let the user understand it, as this suggestion is very misleading.
```suggestion
                f"Reason: {error}") from error
```

conjur/login/login_logic.py
```diff
@@ -0,0 +1,62 @@
...
+
+    This class holds the business logic for populating the
+    netrc configuration details needed to login to Conjur
+    """
+    credentials = None
+
+    def __init__(self, credentials):
+        self.credentials = credentials
+
+    @classmethod
```
abrahamko, line 30: Why static class method?

conjur/login/login_logic.py
```diff
@@ -0,0 +1,62 @@
...
+    def get_api_key(cls, ssl_verify, credential_data, password, loaded_conjurrc):
```
abrahamko, line 31: I think `conjurrc` is a dependency of this command.
Treat it as such, by initializing it with the dependencies, and passing it to both controller and logic in their ctors.

conjur/login/login_logic.py
```diff
@@ -0,0 +1,62 @@
...
+            certificate_path = False
+        # If the cert_file was provided by the user, we will use that cert to login
+        elif loaded_conjurrc['cert_file'] is None and credential_data.machine.startswith("https"):
+            certificate_path = conjur.constants.DEFAULT_CERTIFICATE_FILE
+        else:
+            certificate_path = loaded_conjurrc['cert_file']
+        # pylint: disable=logging-fstring-interpolation
+        logging.debug(f"Attempting to fetch '{credential_data.login}' API key from Conjur")
+        api_key = invoke_endpoint(HttpVerb.GET,
+                                   ConjurEndpoint.LOGIN,
```
abrahamko, line 50: indent correctly

conjur/login/login_logic.py
```diff
@@ -0,0 +1,62 @@
...
+            'account': loaded_conjurrc['account']
+        }
+
+        if ssl_verify is False:
+            certificate_path = False
+        # If the cert_file was provided by the user, we will use that cert to login
+        elif loaded_conjurrc['cert_file'] is None and credential_data.machine.startswith("https"):
+            certificate_path = conjur.constants.DEFAULT_CERTIFICATE_FILE
+        else:
+            certificate_path = loaded_conjurrc['cert_file']
```
abrahamko, line 46: Sort this if/elif/else.
This is very unclear and hard to read.
Simplify it to explicitly demonstrate the usecases you want to handle.

conjur/login/login_logic.py
abrahamko, line 31: ssl_verify can be passed to ctor

conjur/login/login_logic.py
```diff
@@ -0,0 +1,62 @@
...
+class LoginLogic:
+    """
+    LoginLogic
+
+    This class holds the business logic for populating the
+    netrc configuration details needed to login to Conjur
+    """
+    credentials = None
+
+    def __init__(self, credentials):
```
abrahamko, line 27: it seems that the credentials if the `CredentialsFromFile` which can be replaced with similar objects that implement the `save`/`load` interface.
The internal name is, though, misleading.
Call it `credentials_storage` to better explain its purpose.

conjur/login/login_controller.py
abrahamko: Also, isn't it relevant for all the commands? Why is it here?
It should be outside in one place, before starting to handle any command.

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
...
+        self.password = self.get_password()
+        loaded_conjurrc = self.load_conjurrc_data()
+        self.get_api_key(self.ssl_verify, loaded_conjurrc)
+        # pylint: disable=logging-fstring-interpolation
+        logging.debug(f"Saving credentials to '{conjur.constants.DEFAULT_NETRC_FILE}'")
+        self.login_logic.save(self.credential_data)
+        # pylint: disable=logging-fstring-interpolation
+        logging.debug("Credentials successfully written to " \
+                      f"'{conjur.constants.DEFAULT_NETRC_FILE}'")
+        sys.stdout.write("Successfully logged in to Conjur!\n")
```
abrahamko, line 53: Use print.
Also, the exclamation mark is not needed, and looks less professional, IMO (But it's up to the PM to decide).
```suggestion
        print("Successfully logged in to Conjur")
```

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
...
+        Method to fetch the user/host's API key from Conjur which is to be added to the netrc
+        """
+        try:
+            fetched_api_key = self.login_logic.get_api_key(ssl_verify,
+                                                           self.credential_data,
+                                                           self.password,
+                                                           loaded_conjurrc)
+            self.credential_data.password = fetched_api_key
+        except Exception as error:
+            raise RuntimeError("Failed to login. Unable to authenticate with Conjur. " \
```
abrahamko, line 97:
```suggestion
            raise RuntimeError("Failed to login to Conjur. " \
```

conjur/login/login_controller.py
```diff
@@ -0,0 +1,98 @@
...
+        self.credential_data.machine = loaded_conjurrc['appliance_url'] + conjur.constants.NETRC_HOST_PATH
+
+        return loaded_conjurrc
+
+    def get_api_key(self, ssl_verify, loaded_conjurrc):
+        """
+        Method to fetch the user/host's API key from Conjur which is to be added to the netrc
+        """
+        try:
+            fetched_api_key = self.login_logic.get_api_key(ssl_verify,
```
abrahamko, line 91: You can inline the fetched_api_key:
```suggestion
            self.credential_data.password = self.login_logic.get_api_key(ssl_verify,
```

conjur/logout/logout_controller.py
```diff
@@ -0,0 +1,45 @@
...
+    ssl_verify = None
+
+    def __init__(self, ssl_verify, logout_logic):
+        self.ssl_verify = ssl_verify
+        if self.ssl_verify is False:
+            Utils.insecure_message()
```
abrahamko, line 30: See my comment on `login_controller` about this.

conjur/logout/logout_controller.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.logout_logic = logout_logic
+
+    def remove_credentials(self):
+        """
+        Method for removing credentials from user machine
+        """
+        logging.debug("Attempting to log out of Conjur")
+        try:
+            self.logout_logic.remove()
```
abrahamko, line 40: remove what?

conjur/logout/logout_controller.py
```diff
@@ -0,0 +1,45 @@
...
+            logging.debug("Logout successful")
+            sys.stdout.write("Logged out from Conjur\n")
```
abrahamko, line 42: Use print
```suggestion
            print("Logged out from Conjur")
```

conjur/logout/logout_controller.py
```diff
@@ -0,0 +1,45 @@
...
+        except Exception as error:
+            # pylint: disable=raise-missing-from
+            raise Exception(f"Failed to logout. {error}")
```
abrahamko, line 45: Suggestion:
```suggestion
            raise Exception(f"Failed to logout. Reason: {error}")
```

conjur/logout/logout_logic.py
```diff
@@ -0,0 +1,24 @@
...
+class LogoutLogic:
+    """
+    LogoutLogic
+
+    This class holds the business logic for logging out of Conjur
+    """
+    credentials = None
+
+    def __init__(self, credentials):
+        self.credentials = credentials
```
abrahamko, line 18: See my comment about `credentials` in `LoginLogic`

conjur/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    @classmethod
+    def insecure_message(cls):
+        """
+        Method for printing the an insecure warning prompt
+        when user uses --insecure flag
+        """
+        # pylint: disable=line-too-long
+        sys.stdout.write("*" * 100 + "\n")
+        sys.stdout.write("Warning: Running the command with '--insecure' makes your system vulnerable to security attacks!\n")
+        sys.stdout.write("*" * 100 + "\n")
```
abrahamko, line 26: use `print` with multiline string:
```suggestion
        print(f"""{"*" * 100}
                 Warning: Running the command with '--insecure' makes your system vulnerable to security attacks!
                 {"*" * 100}""")
```

abrahamko (changes requested):
All in all, very good work!
It is obvious that you made progress on creating meaningful classes and separating logic to their relevant locations.
I haven't reviewed your tests, @mbenita-Cyberark should get into the details there.

Still, there are some comments that repeated from last review, please pay attention to them.
As a blocker, there's only the password printing.
The rest are still important 😉

conjur/cli.py
```diff
@@ -103,12 +108,14 @@ def run(self, *args):
...
+        init_name = 'init - Initialize Conjur configuration'
+        input_usage = 'conjur [global options] init [options] [args]'
         # pylint: disable=line-too-long
         init_subparser = resource_subparsers.add_parser('init',
                                                         help='Initialize the Conjur configuration',
-                                                        description=self.command_description(init_example, input_usage),
+                                                        description=self.command_description(init_name, input_usage),
                                                         epilog=self.command_epilog('conjur init -a my_org -u https://localhost\t'
                                                                                    'Initializes Conjur configuration and writes to file (.conjurrc)'),
                                                         usage=argparse.SUPPRESS,
```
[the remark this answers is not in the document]
sigalsax: as discussed, we are might evaluate `click`. Regardless, let's save this for a refactor

conjur/cli.py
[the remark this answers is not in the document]
sigalsax: done

conjur/credentials_from_file.py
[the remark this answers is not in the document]
sigalsax: as discussed this change will touch the Client.py which at this time I am don't want to touch because of backwards compatibility

conjur/cli.py
sigalsax: These are keyword arguments <key>=<value>. I cannot do it for one (logout_logic_ and not for the other (ssl_verify) above. I prefer to leave it to be explicit

conjur/cli.py
sigalsax: Glad you asked this question because it means I need to add a comment here. 
In the Ruby CLI, when a user executes a command and they are not logged in, we will prompt the user to log in before we execute the command. This is just good UX because we aren't failing, rather telling them there is a prereq
This is what I am doing here. I am check if the either 1. the netrc file exists or if it exists and is blank. If so, this means the user hasn't logged in and I prompt them to do so.

sigalsax: okay! updating

conjur/cli.py
sigalsax: **much** cleaner. Thanks for the suggestion

conjur/cli.py
sigalsax: They do need to go here because they can accept other arguments. We don't run `conjur login` and execute the command (like we do with variable or policy). It is an interactive command and we need more input.
login / logout are classified as a 'resource' so need to be here. 

If I remove them, I get the help screen

conjur/client.py
sigalsax: added but why do we care to chain the exception if the in the end we get the same message?

conjur/client.py
sigalsax: I do think so because otherwise the ConfigException is not being cascaded up. Changing

conjur/constants.py
sigalsax: 💯 updating to 'CREDENTIAL_HOST_PATH'

conjur/constants.py
sigalsax: netrc only has this logic because netrc is a standard not conjur-unique and not conjur generated file like conjurrc

Nice. I like your suggestion

conjur/credentials_from_file.py
sigalsax: This the UX we want. We want to erase the contents of the file, not remove it

conjur/credentials_from_file.py
sigalsax: I moved it to logout_controller (the caller)

```
logout_controller
try:
    if os.path.exists(DEFAULT_NETRC_FILE) and os.path.getsize(DEFAULT_NETRC_FILE) != 0:
        self.logout_logic.remove()
        logging.debug("Logout successful")
        sys.stdout.write("Logged out from Conjur\n")
    else:
        raise Exception("Please log in")

# creds_from_file
def remove(cls):
    """
    Method that removes netrc data from the file.
    Triggered during a logout
    """
    open(DEFAULT_NETRC_FILE, "w").close()

```

conjur/credentials_from_file.py
sigalsax: > In OOP, you can use object to handle this, instead of static class methods.

Not sure what you mean here

> Its default value could be the const, so you won't have to supply it from outside.

In terms of value. What would this give me? Since I can just change the default netrc path in constants and that's it

conjur/credentials_from_file.py
sigalsax: I added """ and removed the \n
```
with open(self.netrc_path, 'w') as netrc_file:
      netrc_file.write(f"""machine {credential_data.machine}
login {credential_data.login}
password {credential_data.password}
""")
```

conjur/credentials_from_file.py
sigalsax: Wasn't sure so wanted to be explicit but you are right. From what I read [here](https://stackoverflow.com/questions/2404430/does-filehandle-get-closed-automatically-in-python-after-it-goes-out-of-scope), close happens when it breaks from the `with` context

conjur/credentials_from_file.py
sigalsax: I use it to follow suite with the other commands but yes it is overloading/abusing its name. 
I wanted to follow Ruby's `call` which has a call for each CommandClass. Couldn't we do the same here? Would `load_login` be better?

conjur/credentials_from_file.py
sigalsax: 👌

conjur/init/conjurrc_data.py
sigalsax: I don't think it is because default Loader is None

conjur/init/conjurrc_data.py
sigalsax: 👌 why do we want to get the path from outside?

sigalsax: as in why is this better than a const?

conjur/init/init_controller.py
sigalsax: I purposefully didn't want to say "Failed" because we are discovering the account and it is possible that the user is running in OSS so that ability to send a request to /info to discover the account is not an option.

conjur/login/credential_data.py
sigalsax: Stores the API key _but_ the key is 'password' in the netrc

sigalsax: I updated the class attribute to be api_key

conjur/login/credential_data.py
sigalsax: Great catch!!! Is this enough in terms of security? I can just write `***` because the actual value doesnt really need to be printed

conjur/login/credential_data.py
sigalsax: yes it's used in Logout. I will move it

conjur/config.py
sigalsax: Yes line 44 and 54

conjur/credentials_from_file.py
sigalsax: a third party library aka not us (: I added my own logic instead. thnks for the suggestion

conjur/credentials_from_file.py
sigalsax: @mbenita-Cyberark also had this suggestion which is 💯 . This was purposeful because I don't have to change Client because users can still supply this info via client and not directly go through the CLI. In short, this is a bigger change then it seems

conjur/login/login_controller.py
sigalsax: updated this function

conjur/login/login_controller.py
sigalsax: what about init_details? or config?

conjur/init/conjurrc_data.py
sigalsax: love it. Great suggestion

conjur/login/login_controller.py
sigalsax: removing this message because got feedback it is bad UX and annoying

conjur/login/login_logic.py
sigalsax: doesn't use `self` so doesn't need to be a regular method

conjur/login/login_logic.py
sigalsax: You are correct. It is not clear. 
Updating to 
```
        if ssl_verify is False:
            certificate_path = False
        elif credential_data.machine.startswith("https"):
            certificate_path = conjurrc.cert_file
```

conjur/login/login_logic.py
sigalsax: it is passed it. This should be a param because it is available from global scope

conjur/login/login_controller.py
sigalsax: Okay 👌 
I don't want to change to print because I think we should stick to a single way of printing to STDOUT. This change will need to be changed in a bunch of areas which I don't want to be part of this PR

conjur/login/login_controller.py
sigalsax: Will update as per Shuli logs. I think she also mentioned this

conjur/login/login_controller.py
sigalsax: 🥇

conjur/logout/logout_controller.py
sigalsax: removed b/c of PO requirement

conjur/logout/logout_controller.py
sigalsax: remove credentials. will update

conjur/logout/logout_controller.py
sigalsax: see my above comment: https://github.com/cyberark/conjur-api-python3/pull/127#discussion_r550996385

conjur/utils.py
sigalsax: removed

conjur/logout/logout_controller.py
sigalsax: Then it will be `Reason: Please login` which is no bueno

conjur/cli.py
abrahamko: I see.
I have two things to say about this:
1. This `if` must go into the `login` logic. It shouldn't rely here, where it doesn't belong. Add your comment above it to explain why you call `login` every time.
2. IMO, this is not necessarily a better UX. other tools tell you you should log in, and fail, instead of logging you in for each command.
But, if this is the UX guideline that you strive to, why not doing the same for `init`?

conjur/cli.py
abrahamko: Couldn't fully understand, but I trust you.

conjur/client.py
abrahamko: For the debug printing. Inner exception may contain the actual failure data necessary for investigation.

conjur/constants.py
abrahamko: > netrc only has this logic because netrc is a standard not conjur-unique and not conjur generated file like conjurrc

I'm not sure I follow you.
from user's perspective, I don't think there's a difference.
is this the correct UX you wish for Windows users?

sigalsax: yes windows users should have _netrc. macos/linux should have .netrc
See [here](https://www.labkey.org/Documentation/wiki-page.view?name=netrc) and [here]( https://stackoverflow.com/questions/6031214/git-how-to-use-netrc-file-on-windows-to-save-user-and-password)

conjur/cli.py
sigalsax: The logic exists for init as well

FYI @InbalZilberman what do you think about the UX

conjur/credentials_from_file.py
abrahamko: This error is super weird.
Better UX for logout is to fall for no-op in case user is not logged in.
Why telling the user to log in for log out to work?
Also, `Please log in` is not an error.
Error message should contain the info about the failure, in this case: `User is not logged in`.

conjur/logout/logout_controller.py
abrahamko: You shouldn't catch and your own thrown exception in the same try/ecxept.
raise the exception outside, if you need to.
Still, IMO, you don't need to, as log out finished successfully if user is not logged in, and no need to fail it.
Also, `Please log in` is not an error, exception should contain the failure info, so `Reason: ` suggestion apply.

conjur/logout/logout_controller.py
```diff
@@ -0,0 +1,43 @@
...
+        self.ssl_verify = ssl_verify
+        self.logout_logic = logout_logic
+
+    def remove_credentials(self):
+        """
+        Method for removing credentials from user machine
+        """
+        logging.debug("Attempting to log out of Conjur")
+        try:
+            if os.path.exists(DEFAULT_NETRC_FILE) and os.path.getsize(DEFAULT_NETRC_FILE) != 0:
```
abrahamko, line 35: Negate this check and throw exception outside of the try/except clause.
It will be much more readable and make more sense.

conjur/logout/logout_controller.py
```diff
@@ -0,0 +1,43 @@
...
+                self.logout_logic.remove_credentials()
+                logging.debug("Logout successful")
+                sys.stdout.write("Logged out from Conjur\n")
```
abrahamko, line 38: Use `print`:
```suggestion
                print("Logged out from Conjur")
```

conjur/credentials_from_file.py
abrahamko: In terms of value, this class should not rely on const paths, as its purpose is reading and writing the credentials to a file.
The actual path should not impact its behaviour, so it shouldn't depend on a const for that.
It should be a parameter for the object.

conjur/credentials_from_file.py
abrahamko: `retrieve`? it's your call at the end.

conjur/init/conjurrc_data.py
abrahamko: you can still use the const, but keep it as a parameter.
If one day you will want to change it to get the path from the user, you won't need to change much.
at the end, the class purpose is to read/write from the file, regardless of the path, so it shouldn't depend on it.

conjur/init/init_controller.py
abrahamko: So this shouldn't be a warning.
Make it an info.

conjur/login/credential_data.py
abrahamko: You can consult with @moticless if you want, how to correctly handle password.
but for sure, you shouldn't print it to anywhere, unless it is safe.

conjur/config.py
abrahamko: So alias them with `yaml_` prefix.

conjur/credentials_from_file.py
abrahamko: your call.

conjur/login/login_controller.py
abrahamko: I'm all for removing the `conjurrc` concept (together with the `netrc`).
maybe `server_config`?

conjur/login/login_controller.py
abrahamko: 🤣

conjur/login/login_logic.py
abrahamko: Thanks.
Also, `ssl_verify is False` is better read as `not ssl_verify`

conjur/login/login_logic.py
mbenita-Cyberark: ssl_verify leaves in the controller but has no usage instead of passing it to logic.
maybe it should be part of logic instead?
controller should be stateless as possible

conjur/logout/logout_controller.py
mbenita-Cyberark: maybe change the `pelase_login` into` "user not logged in"`

mbenita-Cyberark (commented):
very much readable making my life easy to understand the product

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Ensures that the netrc file is only available its owner
+        os.chmod(self.netrc_path, stat.S_IRWXU)
+
+    def remove_credentials(self):
+        """
+        Method that removes netrc data from the file.
+        Triggered during a logout
+        """
+        open(self.netrc_path, "w").close()
```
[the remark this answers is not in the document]
sigalsax: Yes if the user doesn't have permissions, a Permission denied error will be returned and something like:
```
Exception Failed to log out. {Reason}
```
where that reason is "[Errno 13] Permission denied: <netrc>"

conjur/credentials_from_file.py
```diff
@@ -0,0 +1,68 @@
...
+        accum_netrc = []
+        with open(self.netrc_path, 'r') as netrc_file:
+            for line in netrc_file.readlines():
+                key = line.split()
+                accum_netrc.append(key[1])
+        # pylint: disable=unbalanced-tuple-unpacking
+        _, login_id, api_key = tuple(accum_netrc)
+
+        loaded_netrc['api_key'] = api_key
+        loaded_netrc['login_id'] = login_id
```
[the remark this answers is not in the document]
sigalsax: Oh nice! Being typed will then need to cascade to different functions tho right?

> BTW the dto can have the from_file method and it will make the code easier and testable here

What do you mean here? How would that make it more testable?

conjur/init/conjurrc_data.py
```diff
@@ -6,16 +6,36 @@
...
         self.cert_file = cert_file
         self.plugins = []
 
+    @classmethod
+    def load_from_file(cls, conjurrc_path=DEFAULT_CONFIG_FILE):
+        """
+        Method that loads the conjurrc into an object
+        """
+        with open(conjurrc_path, 'r') as conjurrc:
+            loaded_conjurrc = yaml_load(conjurrc, Loader=YamlLoader)
```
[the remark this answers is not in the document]
sigalsax: in general? or in this project?

conjur/login/login_controller.py
```diff
@@ -0,0 +1,92 @@
...
+This module is the controller that facilitates all login actions
+required to successfully configure the Credentials
+"""
+
+# Builtins
+import getpass
+import logging
+import sys
+
+import conjur.constants
```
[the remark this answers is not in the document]
sigalsax: you are right. I just updated to explicitly state which const I am importing

Thanks!

conjur/login/login_controller.py
```diff
@@ -0,0 +1,92 @@
...
+                raise RuntimeError("Error: Login name is required")
+
+    def get_password(self):
+        """
+        Method to fetch the password from the user attempting to login
+        """
+        if self.user_password is None:
+            # pylint: disable=line-too-long
+            self.user_password = getpass.getpass(prompt=f"Please enter {self.credential_data.login}'s password (it will not be echoed): ")
+            while self.user_password == '':
```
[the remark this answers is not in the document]
sigalsax: I think a space is allowed in a password :o

conjur/credentials_from_file.py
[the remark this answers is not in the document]
mbenita-Cyberark: The exception could be raised in the writing process and the .close() method will never get invoked.
I think it is better to surround it with "with" keyword. It surrounds the code with try catch and finally invoke the close function.
BTW thanks to you I read about it here https://www.geeksforgeeks.org/with-statement-in-python/.

conjur/credentials_from_file.py
mbenita-Cyberark, line 66: Regarding the other comment, sorry I was unclear.
Think you have a DTO called Netrc. it has a from_file method.
the load function here will replace 
```suggestion
        return Netrc.from_file(self.netrc_path)
```
the Netrc file is very easy to test in the code-coverage unitests. while if you have a file parsing issue in the load function it will be harder to detect

conjur/init/conjurrc_data.py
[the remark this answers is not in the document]
mbenita-Cyberark: my question could it be the default Loader and you don't have to give it as a parameter?

conjur/credentials_from_file.py
```diff
@@ -29,40 +31,57 @@ def save(self, credential_data):
...
+                for i, line in enumerate(str(netrc_obj).split('\n')):
+                    if line.strip().startswith('machine') and not i==0:
+                        ret += '\n'
+                    ret += line + '\n'
+                netrc_file.write(ret.replace('\t', ''))
+        else:
+            with open(self.netrc_path, "w+") as netrc_file:
+                credential_data.pretty_print(netrc_file, f"machine {credential_data.machine}")
+                credential_data.pretty_print(netrc_file, f"login {credential_data.login}")
+                credential_data.pretty_print(netrc_file, f"password {credential_data.api_key}")
```
mbenita-Cyberark, line 49: Sorry for the digging😬😬😬😬
why not 
```suggestion
            credential_data.save_to_file(self.netrc_path)
```

conjur/credentials_data.py
```diff
@@ -22,7 +24,34 @@ def __init__(self, machine=None, login=None, password=None):
...
+        # pylint: disable=import-outside-toplevel
+        import netrc
+
+        netrc_obj = netrc.netrc(netrc_path)
+        hosts = netrc_obj.hosts
+        hosts.pop(netrc_data['machine'], None)
+        with open(DEFAULT_NETRC_FILE, 'w') as netrc_file:
+            ret = ""
+            for line in str(netrc_obj).split('\n'):
+                if line.strip().startswith('machine') and not line[0]:
```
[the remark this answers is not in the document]
sigalsax: I wanted to check the first line wrapped it in an enum as we spoke about

conjur/credentials_data.py
```diff
@@ -22,7 +24,34 @@ def __init__(self, machine=None, login=None, password=None):
...
     def __repr__(self):
         return "{'machine': %r, 'login': %r, 'password': ****}" % (self.machine,
                                                                    self.login)
+    @staticmethod
+    def pretty_print(file, str):
+        """
+        Method to write a string to a file
+        """
+        # pylint: disable=redefined-builtin
+        file.write(f"{str}\n")
```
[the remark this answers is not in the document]
sigalsax: So I missed your suggestion on writing the file. What do you say we just do something like
```
            with open(self.netrc_path, "w+") as netrc_file:
                netrc_file.write(f"machine {credential_data.machine}")
                netrc_file.write(f"login {credential_data.login}")
                netrc_file.write(f"password {credential_data.api_key}")

```
that way we get rid of those ugly """

conjur/credentials_from_file.py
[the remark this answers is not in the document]
sigalsax: see above comment about just adding `write`

mbenita-Cyberark (approved):
Very Very goo work

PR #139, opened 2021-01-07
Ensures certificate that is returned from init is used in our tests

sigalsax:
### What does this PR do?
Previously our tests use a pre-configured cert. When sending an init request we were getting 1 cert but couldn't use that cert to make a follow-up requests. This change fixes this by generating a cert and configuring it at the NGINX layer so that we can use the certificate we get back from the server


mbenita-Cyberark (approved):
Great fix

PR #148, opened 2021-01-20
Add host and user implementation and integration tests

### LINKED ISSUE #101: Host and user Commands
## **Feature Overview & Customer Need**

**As a**  Conjur user 
**I would like to** use the conjur cli 
**In order to** update my password 

**As a**  Conjur user or host
**I would like to** use the conjur cli 
**In order to** rotate API Key of a user or a host (it can also be me :))

## COMMANDS

`rotate-api-key`   Rotate a user's or host's API key. If no user/host ID is provided the logged in resource's API Key is rotated.

`change-password`    Update the password of the logged-in user the command is both interactive meaning password is requested from the user and non-interactive using a flag -p

## Help

For each command and subcommand a help is needed see here:
Help for host: https://ljfz3b.axshare.com/#id=paicg6&p=conjur_help__user&g=1
Help for user: https://ljfz3b.axshare.com/#id=paicg6&p=conjur_help__user&g=1

## User messages
please use the following user messages for success scenarios
conjur user rotate-api-key
Success! API key was rotated. New key: 198qmk51cgsa061ghh80x2s56ct99grch1j066xcgnke5431rj3v

conjur host rotate-api-key -i <host_id>
Success! API key was rotated. New key: 3k7k0vb2sf2emjjcphtx14jfvrbbfn8vm3t79mnz19tk2252dsz9wf

conjur user change-password
--request for password--
Success! password was changed.

conjur user change-password -p New-password
Success! password was changed.

All user messages regarding host and user should be reviewed again. Especially error messages. 
If an argument is missing we need to return the help of the command 
## Quality

Make sure we have test coverage of the host commands.  Create test plan and execute accordingly.

## Process logic and Demo

Conjur use with a machine that the Python CLI already been installed and conjur init & login ran

Now the user run the following commands:

`conjur user  rotate-api-key  -i <user_id> `

show the result of rotate-api-key (3v087323qv....ryq38201w9wvx)

`conjur user  change-password 
`
The user is expected to fill in the new password
show response 

`conjur host rotate-api-key  -i <host_id> 
`
show the result of rotate-api-key

## UX Guidelines

Follow guidelines here: https://github.com/cyberark/conjur-api-python3/blob/master/guidelines/python-cli-ux-guidelines.md

## **Documentation**

Please provide enhance documentation in online help and readme

Helpful resources:
user/host epic - https://app.zenhub.com/workspaces/palmtree-5d99d900491c060001c85cba/issues/cyberark/conjur-api-python3/101
## DOD
User (2 commands)
- [x] Integration tests have been added
- [ ] Implementation has been done
- [ ] Log errors/messages have received approval from Inbal + TW

Host (1 command)
- [x] Integration tests have been added
- [ ] Log errors/messages have received approval from Inbal + TW
- [x] Implement the user command & subcommands Conjur Python CLI in all 3 platforms
- [x] Implement the host command & subcommands Conjur Python CLI in all 3 platforms

sigalsax:
### What does this PR do?
This PR adds the following:
- Add implementation for user and host commands
- Add respective integration tests
- Fix a bug where when hopping between users when running the login command, an extra character was being printed. This was happening because of the r+ permissions when opening the netrc file

Added flow:
_user_ 
- rotate-api-key, can rotate personal API key or anothers if logged-in user has privileges on that user
- change-password, change own password
_host_
- rotate-api-key, can rotate a hosts API key if logged-in user has privileges on that host
Resolves #101 

#### Change log
- [x] The CHANGELOG has been updated, or
- [ ] This PR does not include user-facing changes and doesn't require a CHANGELOG update

#### Test coverage

#### Documentation
Will be added


test/test_integration_resource.py
```diff
@@ -0,0 +1,182 @@
...
+        attempt_to_rotate_user_key = self.invoke_cli(self.cli_auth_params,
+               ['user', 'rotate-api-key', '-i', 'someuser'], exit_code=1)
+        self.assertIn("401 Client Error", attempt_to_rotate_user_key)
+
+        # Verify that a user cannot rotate an admin's API key
+        attempt_to_rotate_admin_key = self.invoke_cli(self.cli_auth_params,
+               ['user', 'rotate-api-key', '-i', 'admin'], exit_code=1)
+        self.assertIn("500 Server Error", attempt_to_rotate_admin_key)
+
+    def test_user_rotate_api_key_without_flag_returns_error(self):
```
eladkug, line 125: Some of the tests missing the @integration_test decorator.

test/test_integration_resource.py
```diff
@@ -0,0 +1,182 @@
...
+        with redirect_stderr(self.capture_stream):
+            self.invoke_cli(self.cli_auth_params,
+                            ['user', 'rotate-api-key', 'someinput'], exit_code=1)
+        self.assertIn('Error', self.capture_stream.getvalue())
```
eladkug, line 129: maybe we need to be more specific here ?

conjur/cli.py
```diff
@@ -414,6 +444,27 @@ def run_action(resource, args):
...
+        elif resource == 'user':
+            credentials = CredentialsFromFile()
+            user_logic = UserLogic(ConjurrcData, credentials, client, resource)
+
+            if args.action == 'rotate-api-key':
+                user_resource_data = UserResourceData(action=args.action, id=args.id, new_password=None)
+                user_controller = UserController(user_logic=user_logic,
+                                                user_resource_data=user_resource_data)
+                user_controller.rotate_api_key()
+            else:
```
tzheleznyak, line 456: use else if with the action name

conjur/host/host_resource_data.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, **arg_params):
+        self.action = arg_params['action']
+        self.host_to_update = arg_params['host_to_update']
+
+    def __repr__(self):
+        result = []
+        # pylint: disable=multiple-statements
+        if self.action: result.append(f"'action': '{self.action}'")
+        if self.host_to_update: result.append(f"'host': '{self.host_to_update}'")
+        return '{'+', '.join(result) + '}'
```
tzheleznyak, line 24: fstring

conjur/user/user_controller.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def change_password(self):
+        """
+        Method that calls to change password to perform the change
+        """
+        self.prompt_for_password()
+        try:
+            self.user_logic.change_password(self.user_resource_data.new_password)
+        except requests.exceptions.HTTPError as http_error:
+            raise requests.exceptions.HTTPError("Invalid password. " \
```
tzheleznyak, line 41: Create an object for it  `PasswordNotComplexEnough`

conjur/user/user_logic.py
```diff
@@ -0,0 +1,88 @@
...
+
+# Internals
+from conjur.credentials_data import CredentialsData
+
+class UserLogic:
+    """
+    UserLogic
+
+    This class holds the business logic for handling
+    user activity
```
tzheleznyak, line 20: put this two lines in one line it is very short

conjur/user/user_logic.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        Method to trigger two types of API key actions.
+        1. The user can either rotate their own API key or
+        2. Rotate another users API key
+        """
+        if resource_to_update is not None:
+            logging.debug(f"Rotating API key for '{resource_to_update}'")
+            new_api_key = self.rotate_anothers_api_key(resource_to_update)
+            logging.debug(f"API key for '{resource_to_update}' successfully rotated")
+        else:
```
tzheleznyak, line 38: From UX perspective are we sure we want to rotate if the user dont specify anything ? Maybe we want him to specifiy it implicitly? Because if he use it without flags maybe he won't want to rotate his own api key?

conjur/user/user_resource_data.py
```diff
@@ -0,0 +1,26 @@
...
+        self.id = arg_params['id']
+        self.new_password = arg_params['new_password']
+
+    def __repr__(self):
+        result = []
+        # pylint: disable=multiple-statements
+        if self.action: result.append(f"'action': '{self.action}'")
+        if self.action == 'rotate-api-key': result.append(f"'id': '{self.id}'")
+        if self.action == 'change-password': result.append("'new_password': '****'")
+        return '{'+', '.join(result) + '}'
```
tzheleznyak, line 26: fstring

conjur/user/user_controller.py
```diff
@@ -0,0 +1,52 @@
...
+                                                "Please verify that it has the correct password complexity") from http_error
+        logging.debug(f"Password successfully changed")
+        sys.stdout.write('Password has been successfully changed\n')
+
+    def prompt_for_password(self):
+        if self.user_resource_data.new_password is None:
+            # pylint: disable=line-too-long
+            self.user_resource_data.new_password = getpass.getpass(prompt="Please enter the new password (it will not be echoed): ")
+            while self.user_resource_data.new_password == '':
```
tzheleznyak, line 50: Move password validation to another function that you call it here for case we want to add more logic

tzheleznyak: Password complexity check for example

tzheleznyak (changes requested):
Left some comments

test/test_integration_resource.py
```diff
@@ -0,0 +1,182 @@
...
+        return self.cli_auth_params
+
+    def setUp(self):
+        self.setup_cli_params({})
+        # Need to configure the CLI and login to perform further commands
+        Utils.setup_cli(self)
+        self.invoke_cli(self.cli_auth_params,
+                       ['policy', 'replace', 'root', self.environment.path_provider.get_policy_path("resource")])
+
+    @integration_test
```
eladkug, line 42: Add section for test *** TESTS ***

test/test_integration_resource.py
```diff
@@ -0,0 +1,182 @@
...
+            self.invoke_cli(self.cli_auth_params,
+                            ['user', 'rotate-api-key', 'someinput'], exit_code=1)
+        self.assertIn('Error', self.capture_stream.getvalue())
+
+    def test_user_rotate_api_key_flag_without_value_returns_error(self):
+        with redirect_stderr(self.capture_stream):
+            self.invoke_cli(self.cli_auth_params,
+                            ['user', 'rotate-api-key', '-i'], exit_code=1)
+
+        self.assertIn('Error', self.capture_stream.getvalue())
```
eladkug, line 136: maybe we need to be more specific here ?

test/test_integration_resource.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @integration_test
+    def test_host_rotate_api_key_with_host_provided_rotates_host_api_key(self):
+        output = self.invoke_cli(self.cli_auth_params,
+               ['host', 'rotate-api-key', '-i', 'somehost'])
+        self.assertIn("API key for 'somehost' was successfully rotated", output)
+
+    def test_host_rotate_api_key_without_flag_returns_error(self):
+        output = self.invoke_cli(self.cli_auth_params,
+                        ['host', 'rotate-api-key', '-i', 'someinput'], exit_code=1)
```
eladkug, line 175: you need to remove the flag from here

eladkug (dismissed):
Left some comments. 
I didnt see test_unprivileged_host_cannot_rotate_anothers_api_key as part of test integration.

test/test_integration_resource.py
sigalsax: why do I need to remove the flag?

test/test_integration_resource.py
sigalsax: you are correct. updated

sigalsax:
@eladkug 
> I didnt see test_unprivileged_host_cannot_rotate_anothers_api_key as part of test integration.

Yes this test is not relevant because you cannot login as a host so you cannot attempt to rotate another hosts API key

test/test_integration_resource.py
sigalsax: updated 👌

conjur/cli.py
sigalsax: updated

test/test_integration_resource.py
eladkug: The name of the test **without_flag_returns_error**

conjur/user/user_logic.py
sigalsax: This is the same ux as in Ruby but it is question for @sharonr78 why we want to keep this.

sharonr78: Hi @sigalsax and @tzheleznyak,
The suggested UX is to prompt for a username or host to rotate their key, but when used without flags than it is implicit.
For instance:
`conjur user rotate-api-key -i joe` will rotate the key for user joe, while,
`conjur user rotate-api-key` will prompt the logged-in user to specify a name for api key rotation. It can be his own name, or another name.
- Checkout this commands mapping for a reference: https://ljfz3b.axshare.com/#id=c63xjv&p=commands_matrix___rules&g=1

@InbalZilberman please confirm

test/test_integration_resource.py
sigalsax: updated

conjur/user/user_logic.py
sigalsax: These described flows are what is being implemented. @tzheleznyak  is suggesting that `conjur use rotate-api-key` is too intrusive (correct me if I am wrong Tamir) and a user could run `conjur user rotate-api-key` and expect to see the help screen for example for further guidance

conjur/cli.py
```diff
@@ -339,48 +367,75 @@ def handle_logout_logic(cls, ssl_verify=True):
...
+        """
+        credentials = CredentialsFromFile()
+        user_logic = UserLogic(ConjurrcData, credentials, client, resource)
+        if args.action == 'rotate-api-key':
+            user_resource_data = UserResourceData(action=args.action,
+                                                  id=args.id,
+                                                  new_password=None)
+            user_controller = UserController(user_logic=user_logic,
+                                             user_resource_data=user_resource_data)
+            user_controller.rotate_api_key()
```
mbenita-Cyberark, line 421: better to extract to private function.. same for other else_ifs here

mbenita-Cyberark: btw if you think it will make the class noisy we can create class handle_user_logic. will look like this

```suggestion
class handle_user_logic:
    def __init__():
         credentials = CredentialsFromFile()
        user_logic = UserLogic(ConjurrcData, credentials, client, resource)
    def rotate_api_key():
        user_resource_data = UserResourceData(action=args.action,
                                                  id=args.id,
                                                  new_password=None)
            user_controller = UserController(user_logic=user_logic,
                                             user_resource_data=user_resource_data)
            user_controller.rotate_api_key()
       def change_password():
           user_resource_data = UserResourceData(action=args.action,
                                                  id=None,
                                                  new_password=args.password)
            user_controller = UserController(user_logic=user_logic,
                                             user_resource_data=user_resource_data)
            user_controller.change_password()
       
```

conjur/client.py
```diff
@@ -213,3 +213,22 @@ def update_policy_file(self, policy_name, policy_file):
         Replaces a file-based policy defined in the Conjur instance
         """
         return self._api.update_policy_file(policy_name, policy_file)
+
+    def rotate_anothers_api_key(self, resource, resource_to_rotate):
+        """
+        Rotates a API keys and returns new API key
+        """
+        return self._api.rotate_anothers_api_key(resource, resource_to_rotate)
```
mbenita-Cyberark, line 221: do we want to add a client side validation? (permission wise.) just asking

conjur/credentials_data.py
```diff
@@ -24,20 +27,24 @@ def __init__(self, machine=None, login=None, password=None):
...
         netrc_obj = netrc.netrc(netrc_path)
         hosts = netrc_obj.hosts
         hosts.pop(netrc_data['machine'], None)
 
         netrc_obj = netrc.netrc(netrc_path)
         netrc_obj.hosts.pop(netrc_data['machine'], None)
+        cls.build_netrc_from_instance(netrc_obj)
+
+    @classmethod
+    def build_netrc_from_instance(cls, netrc_obj):
```
mbenita-Cyberark, line 44: why is it a class method and not a util method in the UTILS class?

conjur/host/host_controller.py
```diff
@@ -0,0 +1,43 @@
...
+        Method that distinguishes between the type of actions
+        and facilitates the calls accordingly
+        """
+        self.prompt_for_host_id()
+        new_api_key = self.client.rotate_anothers_api_key(resource,
+                                                          self.host_resource_data.host_to_update)
+        sys.stdout.write(f"API key for '{self.host_resource_data.host_to_update}' " \
+                        f"was successfully rotated. New API key is: {new_api_key}\n")
+
+    def prompt_for_host_id(self):
```
mbenita-Cyberark, line 33: general.. incase the function can not raise a prompt (case host_to_update is None) should add the if_needed suffix prompt_for_host_id-> prompt_for_host_id_if_needed

conjur/host/host_controller.py
```diff
@@ -0,0 +1,43 @@
...
+
+class HostController():
+    """
+    HostController
+
+    This class represents the Presentation Layer for the User command.
+    """
+    def __init__(self, client, host_resource_data):
+        self.client=client
+        self.host_resource_data=host_resource_data
```
mbenita-Cyberark, line 20: for the refactoring part. should add input validation some way in the flow (host is not emty string and more).
maybe constructor of controllers is a good place to validate

conjur/user/user_logic.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Method to call the client to change the logged in user's password
+        """
+        loaded_credentials = self.extract_credentials_from_credential_store()
+        resource_to_update = loaded_credentials['login_id']
+        logging.debug(f"Changing password for '{resource_to_update}'")
+        # pylint: disable=line-too-long
+        return self.client.change_password(resource_to_update, loaded_credentials['api_key'], new_password)
+
+    def extract_credentials_from_credential_store(self):
```
mbenita-Cyberark, line 73: goo encapsulation. all conjurrc and netrc action are better to be encapsulate as we might change the way we save credential in the future

conjur/user/user_logic.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # pylint: disable=logging-fstring-interpolation
+    def rotate_api_key(self, resource_to_update):
+        """
+        Method to trigger two types of API key actions.
+        1. The user can either rotate their own API key or
+        2. Rotate another users API key
+        """
+        loaded_credentials = self.extract_credentials_from_credential_store()
+        user_from_credential_store = loaded_credentials['login_id']
```
mbenita-Cyberark, line 36: login_id  is better to be strong typed and not extracted as a key from loaded_credentials.
if so i suggest put 'login_id' as a const somewhere and not as a magic string

conjur/user/user_logic.py
```diff
@@ -0,0 +1,100 @@
...
+        if resource_to_update == user_from_credential_store:
+            raise Exception("Error: To rotate the API key of the currently logged-in user "\
+                            " use this command without any flags or options")
```
mbenita-Cyberark, line 39: I think that user instructions should be created in the controller and the logic should throw logic exceptions. also it is recommended to be as much as specific as you can when throwing an error so InvalidOperationException is better. if you can't find the type of exception you want to throw you can allways create one. it will be good if we have exceptions module for the CLI

conjur/user/user_logic.py
```diff
@@ -0,0 +1,100 @@
...
+            # if the user does not provide a user to rotate their API key,
+            # their own API key will be rotated
+            logging.debug(f"Rotating API key for '{user_from_credential_store}'")
+            new_api_key = self.rotate_personal_api_key(user_from_credential_store,
+                                                       loaded_credentials['api_key'])
+
+            # Update the new rotated API for the logged-in user
+            logging.debug("Updating credential store with new API key " \
+                          f"for '{user_from_credential_store}'")
+            self.update_api_key_in_credential_store(user_from_credential_store,
```
mbenita-Cyberark, line 56: incase an error occurs between setting the new api_key and update it in the store (again good encapsulation). we are in an unstable situation where the api key at the server is different from the store. maybe we should alert the user? or try to fetch the key from the server 
```suggestion
    try:
            new_api_key = self.rotate_personal_api_key(user_from_credential_store,
                                                       loaded_credentials['api_key'])

            # Update the new rotated API for the logged-in user
            logging.debug("Updating credential store with new API key " \
                          f"for '{user_from_credential_store}'")
            self.update_api_key_in_credential_store(user_from_credential_store,
     except:
         reRaise exception
    finally:
        #get api key from server and store it on the store
```

mbenita-Cyberark (commented):
Sorry for not going the whole PR.. looks very good

conjur/api.py
```diff
@@ -285,6 +284,46 @@ def update_policy_file(self, policy_id, policy_file):
 
         return self._load_policy_file(policy_id, policy_file, HttpVerb.PATCH)
 
+    def rotate_anothers_api_key(self, resource, resource_to_rotate):
```
abrahamko, line 287: suggestion:
```suggestion
    def rotate_other_api_key(self, resource, resource_to_rotate):
```

conjur/api.py
```diff
@@ -285,6 +284,46 @@ def update_policy_file(self, policy_id, policy_file):
...
+        """
+        This method is used to rotate a user/host's API key
```
abrahamko, line 289:
```suggestion
        This method is used to rotate a user/host's API key that is not the current user.
        To rotate API key of current user use rotate_personal_api_key
```

conjur/api.py
```diff
@@ -285,6 +284,46 @@ def update_policy_file(self, policy_id, policy_file):
...
+        """
+        # Attach the resource type (user or host)
+        full_resource_id=':'.join([resource, resource_to_rotate])
```
abrahamko, line 292: do you validate resource is valid? that is, user/host?

conjur/api.py
```diff
@@ -285,6 +284,46 @@ def update_policy_file(self, policy_id, policy_file):
...
+        This method is used to rotate a personal API key
+        """
+        response = invoke_endpoint(HttpVerb.PUT, ConjurEndpoint.ROTATE_API_KEY,
+                                   self._default_params,
+                                   api_token=self.api_token,
+                                   auth=(resource_to_rotate, current_password),
+                                   ssl_verify=self._ssl_verify).text
+        return response
+
+    def change_password(self, resource_password_to_change, current_password, new_password):
```
abrahamko, line 314: maybe `change_personal_password`? similar to `rotate_personal_api_key`?
IMO it makes it clearer.

conjur/api.py
```diff
@@ -285,6 +284,46 @@ def update_policy_file(self, policy_id, policy_file):
...
+            'role': full_resource_id
+        }
+        response = invoke_endpoint(HttpVerb.PUT, ConjurEndpoint.ROTATE_API_KEY,
+                                   self._default_params,
+                                   api_token=self.api_token,
+                                   ssl_verify=self._ssl_verify,
+                                   query=query_params).text
+        return response
+
+    def rotate_personal_api_key(self, resource_to_rotate, current_password):
```
abrahamko, line 303: isn't `resource_to_rotate` always equal `user`?
Why pass it if you know it?

conjur/api.py
abrahamko, line 314: isn't `resource_password_to_change` always equal user?
Why pass it if you know it?

conjur/credentials_data.py
abrahamko: I say the opposite, why is it a class method and not an instance method?
I must say I'm not a big fan of this class structure.
Could be better encapsulated IMO.
As in, let the object handle itself internally, and only expose what you need.
This mixture of static/class methods and instance methods could be better designed.

conjur/client.py
```diff
@@ -119,7 +119,7 @@ def __init__(self,
                 credentials = CredentialsFromFile(DEFAULT_NETRC_FILE)
                 loaded_netrc = credentials.load(loaded_config['url'])
             except netrc.NetrcParseError as netrc_error:
-                raise Exception("Error: netrc in an invalid format." \
+                raise Exception("Error: netrc in an invalid format. " \
```
abrahamko, line 122: Sorry to say, but this doesn't look like correct english... ;)
Maybe:
```suggestion
                raise Exception("Error: netrc is in an invalid format. " \
```

conjur/client.py
abrahamko: this is a long winding road.
Overloaded server could benefit from such approach (PVWA does this to Vault), but I think this needs to be added as an explicit performance feature, across all out major clients, and only if it is proven to improve server performance significantly.

conjur/login/login_controller.py
```diff
@@ -25,6 +25,8 @@ class LoginController:
 
     def __init__(self, ssl_verify, user_password, credential_data, login_logic):
         self.ssl_verify = ssl_verify
+        # We check if the user supplied --insecure because
+        # this command does not go through the Client
```
abrahamko, line 29: Sorry, I didn't get the prupose of this comment nor its meaning...
Could you better explain?

abrahamko (commented):
Well done, @sigalsax!
I have one general comment.
It looks like you combined into this PR multiple separated tasks.
It's good practice to create a PR for each, for both easier review, and cleaner git history. Then, each commit represents exactly the change it should and does not hide more changes internally.

conjur/user/user_controller.py
sigalsax: a Conjur CLI-specific exception as in?

conjur/cli.py
sigalsax: this would then mean I would need to do this for all other handle_x_logic in this CLI class which is not what I want to get into now. I will open a card for a better design

conjur/client.py
sigalsax: Agreed ^ 
I don't think there is much we can do specifically for rotate_anothers_api_key beyond making sure the user exists but don't think it would be worth the effort here. We could for example, on `change-password` action have some validation that a password meets the Conjur password complexity and allow the user another opportunity to input their new password without failing the request (as we are doing now). I also think that this is a widespread effort that is not unique to this PR that requires more thought. I will open a ticket for this effort because it can be applied in a lot of palces.

conjur/host/host_controller.py
sigalsax: If it is empty i prompt them to enter in the host id. If they anyway enter a empty host, then it will fail. So that is some sort of validation no?

conjur/api.py
sigalsax: resource_to_rotate is not the actual resource but the name of the resource to rotate, for example (somehost, someuser). Fact that this is confused makes me want to change the name. I will change it to `id_of_logged_in_user`

conjur/api.py
sigalsax: same applies here. https://github.com/cyberark/conjur-api-python3/pull/148#discussion_r563290330

it should be the name of the logged in user not `user`

conjur/login/login_controller.py
sigalsax: Client (client.py) when initialized has this check (if not ssl_verify, print an insecure message). For this command (init and login/logout) we don't initialize the client because we don't want to send an request to Conjur yet. Therefore, we need to have this check out here

sigalsax: Client (client.py) when initialized has this check (if not ssl_verify, print an insecure message). For these commands (init and login/logout) we don't initialize the client because we don't have the needed information to be successful yet. Therefore, we need to have this check out here

conjur/api.py
sigalsax: I don't because resource comes from the CLI which is args.action (user or host). Now that I think about it, we are also the SDK 🙃  which doesnt have this information. So I will validate! Adding:
```
if resource != 'user' and resource != 'host':
     raise Exception("Error: Invalid resource type")
```

conjur/user/user_logic.py
sigalsax: do you have an example of what you are suggesting? B/c the value of the const depends on the above line,
`loaded_credentials = self.extract_credentials_from_credential_store()`

conjur/user/user_logic.py
sigalsax: Agreed! but this logic is appropriate only here because we have extracted who the logged in user is. How would you get around this?

I agree I should make this a CLI-specific error. Will do

sigalsax: how would you call this custom error?
I was thinking `RotateApiKeyUserLoggedInUserSameAsFlag` but that is very very long

conjur/credentials_data.py
sigalsax: @abrahamko it is not an instance method because it has no use for 'self' here. I redesigned this class so that credentials_data will be a POJO and credentials_from_file is where all the data manipulation is happening (updating the netrc, removing netrc data)

conjur/user/user_logic.py
sigalsax: Correct it would lead to an unstable situation and needs to be addressed! One thing though, I am not understanding how the above suggestion is overcoming this instability because in the end we are just making the same request twice
1. Rotate API key
2. Attempt to store
3. Fail
4. Rotate API key (again)
5. Attempt to store (again)
There is no way to get the API key again, only rotate.

How do you feel about the following flow
1. Rotate API key
2. Attempt to store
3. Fail
4. Return saying that the API key was rotated and it is AX12 but something went wrong and is in an unstable state.

conjur/user/user_controller.py
sigalsax: adding

conjur/user/user_controller.py
sigalsax: adding

conjur/credentials_data.py
mbenita-Cyberark: Sigal I was thinking this is a different class this indeed seems to be the place for this kind of function. sorry

conjur/host/host_controller.py
mbenita-Cyberark: yep.. you have it all covered.

conjur/user/user_logic.py
mbenita-Cyberark: No example, just it's preferred to not use magic strings, and rather to const it before
https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings

conjur/user/user_logic.py
mbenita-Cyberark: for the first comment - throw here and catch in the controller. there you can set the user message

for the second - i think this is too specific I don't think you will use `RotateApiKeyProvidedUserIsSameAsLoggedIn ` 
 somewhere else (if you think I'm wrong than it is indeed good exception).
 this looks like invalidOperation so what about `invalidOperationException`? this shows what is the reason for the exception + you can specify the exact reason in the string inside.

now the controller can have a 

`try: .. 
except invalidOperationException: handle specific error here and show the user the desired message 
except Exception: handle general error that you didn't see coming`

conjur/user/user_logic.py
mbenita-Cyberark: sounds good. 
one thing that i don't understand. if the rotate failed than we can't perform login and get the api key.
I think that the error message to the user should be something like "rotate failed please login again to get the currect api-key."
but if we expect the user to login afterwards than we can to it ourselves in the `except` or `finally` stage as part of the function error handling .

conjur/user/user_logic.py
sigalsax: > for the first comment - throw here and catch in the controller. there you can set the user message

if I am understanding, you are suggesting I throw the InvalidOperation from the logic and intercept and rethrow from the controller? So it looks like:
```
# logic.py

raise InvalidOperation

# controller.py
try:
..
except InvalidOperation
...
except Exception
...

File: conjur/user/user_logic.py (hunk 21 above)
Comment (sigalsax) [reply conjur/user/user_logic.py:57]: The general statement is `Failed to execute command. Reason: 401 Client Error: Unauthorized for url: https://localhost/authn/cucumber/adfmin/authenticate` `Failed to execute command` is general enough so I dont think we need to say rotate failed because we failed on authentication

File: conjur/user/user_logic.py (hunk 20 above)
Comment (mbenita-Cyberark) [reply conjur/user/user_logic.py]: exactly

File: conjur/host/host_controller.py [hunk 29]
```diff
@@ -0,0 +1,44 @@
...
+        """
+        Method to prompt the user to enter the host id of the
+        host whose API key they want to rotate
+        """
+        if self.host_resource_data.host_to_update is None:
+            # pylint: disable=line-too-long
+            self.host_resource_data.host_to_update = input("Enter the host id to rotate its API key: ").strip()
+            if self.host_resource_data.host_to_update  == '':
+                # pylint: disable=raise-missing-from
+                raise RuntimeError("Error: Host id is required")
```
tzheleznyak, line 44: Create your own exeption object like MissingHostException

conjur/login/login_controller.py
```diff
@@ -25,6 +25,10 @@ class LoginController:
 
     def __init__(self, ssl_verify, user_password, credential_data, login_logic):
         self.ssl_verify = ssl_verify
+        # For init/login commands, the client (client.py) is not initialized
```
tzheleznyak, line 28: move this comment above line `self.ssl_verify = ssl_verify`
and put them between """ and """

conjur/logout/logout_logic.py
```diff
@@ -17,8 +17,8 @@ class LogoutLogic:
     def __init__(self, credentials):
         self.credentials = credentials
 
-    def remove_credentials(self, conjurrc_appliance_url):
+    def remove(self, conjurrc_appliance_url):
```
tzheleznyak, line 20: i think privious name is better. Remove without anything dont give me much information

conjur/user/user_controller.py
```diff
@@ -0,0 +1,78 @@
...
+from conjur.errors import InvalidPasswordComplexity, InvalidOperation
+
+
+class UserController():
+    """
+    UserController
+
+    This class represents the Presentation Layer for the User command.
+    """
+    password_complexity_constraints = "The password must contain at least 12 characters: " \
```
tzheleznyak, line 25: SHOULD be capital case letters because it is const

tzheleznyak: Also it belongs to class and not to instance itself so maybe use it without anything or with `UserController.PASSWORD_COMPLEXITY_CONSTRAINS`

tzheleznyak: Also maybe adding the word message to the names can help but this is up to you to decide

tzheleznyak (changes requested):
Left some small comments

conjur/login/login_controller.py
sigalsax: you got it!

conjur/user/user_controller.py
sigalsax: I actually think it would be better in constants
moving, capitalizing,  and adding message to the end

tzheleznyak (approved):
Great job @sigalsax

conjur/api.py
abrahamko: You can do it using `in` operator in python, makes it more readable.

conjur/api.py
abrahamko: I think `user_name` is enough, but it's your call.
I just don't like using `id` as it is not clear.

conjur/api.py
```diff
@@ -285,6 +284,56 @@ def update_policy_file(self, policy_id, policy_file):
 
         return self._load_policy_file(policy_id, policy_file, HttpVerb.PATCH)
 
+    def rotate_other_api_key(self, resource, resource_to_rotate):
```
abrahamko, line 287: I suggest better naming (especially important in signatures)
```suggestion
    def rotate_other_api_key(self, resource_type, resource_name):
```

And this leads me to thing that they can be combined into one type `Resource` that also has the method `id` that does the `':'.join([resource, resource_to_rotate])`.
You could even take it one level further and derive `Host` and `User` from it, which will have the type defined as const for each instance of the class.

conjur/user/user_controller.py
```diff
@@ -0,0 +1,82 @@
...
+            # pylint: disable=line-too-long
+            resource_to_update, new_api_key = self.user_logic.rotate_api_key(self.user_resource_data.user_id)
+            sys.stdout.write(f"Successfully rotated API key for '{resource_to_update}'. " \
+                             f"New API key is: {new_api_key}\n")
+        except InvalidOperationException as invalid_operation:
+            raise InvalidOperationException(f"Error: {invalid_operation}") from invalid_operation
+        except OperationNotCompletedSuccessfullyException as operation_not_completed:
+            # pylint: disable=line-too-long
+            raise OperationNotCompletedSuccessfullyException("Error: Failed to run command to completion leaving " \
+                                                             "the CLI in an unstable state. Log in again or " \
```
abrahamko, line 45: as a rule of thumb, errors should not contain instructions.
If you want to write to stdout, do it.
I don't see the value in wrapping an exception with the same exception without adding important information.

conjur/user/user_logic.py
abrahamko: Why don't you use an object with fields instead of map? Then you would have `credentials.login_id`

conjur/user/user_logic.py
```diff
@@ -0,0 +1,108 @@
...
+        Method to trigger two types of API key actions.
+        1. The user can either rotate their own API key or
+        2. Rotate another users API key
+        """
+        loaded_credentials = self.extract_credentials_from_credential_store()
+        user_from_credential_store = loaded_credentials['login_id']
+        if resource_to_update == user_from_credential_store:
+            raise InvalidOperationException("To rotate the API key of the currently logged-in user "\
+                                   "use this command without any flags or options")
+        if resource_to_update is not None:
```
abrahamko, line 42: I would check for `is None` first, and leave the `else` to refer to `is not None`.
```
if a is None:
  ...
else:
  ...
```
is more readable than:
```
if a is not None:
  ...
else:
  ...
```
That is b/c `else` after `not None` is double negation, and you should avoid it as much as possible.

conjur/user/user_logic.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # pylint: disable=logging-fstring-interpolation,line-too-long
+    def rotate_api_key(self, resource_to_update):
+        """
+        Method to trigger two types of API key actions.
+        1. The user can either rotate their own API key or
+        2. Rotate another users API key
+        """
+        loaded_credentials = self.extract_credentials_from_credential_store()
+        user_from_credential_store = loaded_credentials['login_id']
```
abrahamko, line 38: I would prefer a more concise name other than `user_from_credential_store`.
Maybe `current_user_name` or `logged_in_user_name`.
Adding the `name` suffix is important to denote its content, unless you will replace it with my suggestion for `User` class instance.

conjur/api.py
sigalsax: the api is called by the client whose signature looks like:

```
def rotate_other_api_key(self, resource, resource_to_rotate):
```
Those that use our SDK call `client.rotate_other_api_key('user', 'someuser')`, so we cannot pass in some Resource object.

I can do work inside that function to join `':'.join([resource, resource_to_rotate])` but don't understand what is wrong with the naming `resource_type` / `resource_name`

conjur/user/user_controller.py
sigalsax: I want to perc this exception up to the `cli.py` because there is the following catch
```
except Exception as error:
    logging.debug(traceback.format_exc())
    sys.stdout.write(f"{str(error)}\n")
    sys.exit(1)
```
which takes care of printing the backtrace in debug mode and formatting.

conjur/user/user_logic.py
sigalsax: Totally agree but wanted to handle the shorter logic first so it more readable. So it becomes a question of either short logic first or double negative. Do you still recommend going with `if a is None: ...long logic else: short logic`

conjur/api.py
abrahamko: > Those that use our SDK call client.rotate_other_api_key('user', 'someuser'), so we cannot pass in some Resource object.

Don't be so negative and say "cannot". I can suggest some options how to handle this, if you wish.
Anyway, it's up to you.

> don't understand what is wrong with the naming resource_type / resource_name

If I understand correctly, you accepted my rename suggestion, and question the need of an object to wrap them.
The need comes from the fact that these params are related and may repeat multiple times throughout the code.
Moreover, there's more logic specially related to them, so it is clear to me that they deserve a class of their own.

conjur/user/user_controller.py
abrahamko: I still think my suggestion stands:
> If you want to write to stdout, do it.

And so is my comment:
> I don't see the value in wrapping an exception with the same exception without adding important information.

So you can throw the same exception (no wrapping), just write to stdout whatever that you want to write.

conjur/api.py
sigalsax: as we spoke about, I will update the param as a resource that consists of resource_type and resource_name with id as a function to make that join (:

conjur/user/user_controller.py
sigalsax: as we spoke about, I am throwing this message in the actual exception body

PR #152, opened 2021-01-26
Implements additional feedback from user/host commands

sigalsax:
### What does this PR do?
Additional feedback was received on the user/host commands. For the sake of an easier review process, I am creating a new PR that addresses the feedback

### DIFF
--- a/conjur/api.py
+++ b/conjur/api.py
@@ -284,18 +284,17 @@ class Api():
 
         return self._load_policy_file(policy_id, policy_file, HttpVerb.PATCH)
 
-    def rotate_other_api_key(self, resource, resource_to_rotate):
+    def rotate_other_api_key(self, resource_obj):
         """
         This method is used to rotate a user/host's API key that is not the current user.
-        To rotate API key of current user use rotate_personal_api_key
+        To rotate API key of the current user use rotate_personal_api_key
         """
-        if resource not in ('user', 'host'):
+        if resource_obj.resource_type not in ('user', 'host'):
             raise Exception("Error: Invalid resource type")
 
         # Attach the resource type (user or host)
-        full_resource_id=':'.join([resource, resource_to_rotate])
         query_params = {
-            'role': full_resource_id
+            'role': resource_obj.full_id()
         }
         response = invoke_endpoint(HttpVerb.PUT, ConjurEndpoint.ROTATE_API_KEY,
                                    self._default_params,
@@ -304,13 +303,10 @@ class Api():
                                    query=query_params).text
         return response
 
-    def rotate_personal_api_key(self, resource, logged_in_user, current_password):
+    def rotate_personal_api_key(self, logged_in_user, current_password):
         """
         This method is used to rotate a personal API key
         """
-        if resource != 'user':
-            raise Exception("Error: Invalid resource type")
-
         response = invoke_endpoint(HttpVerb.PUT, ConjurEndpoint.ROTATE_API_KEY,
                                    self._default_params,
                                    api_token=self.api_token,
@@ -318,13 +314,10 @@ class Api():
                                    ssl_verify=self._ssl_verify).text
         return response
 
-    def change_personal_password(self, resource, logged_in_user, current_password, new_password):
+    def change_personal_password(self, logged_in_user, current_password, new_password):
         """
         This method is used to change own password
         """
-        if resource != 'user':
-            raise Exception("Error: Invalid resource type")
-
         response = invoke_endpoint(HttpVerb.PUT, ConjurEndpoint.CHANGE_PASSWORD,
                                    self._default_params,
                                    new_password,
--- a/conjur/api.py
+++ b/conjur/api.py
@@ -18,6 +18,9 @@ from conjur.http_wrapper import HttpVerb, invoke_endpoint
 
 
 # pylint: disable=too-many-instance-attributes
+from conjur.resource import Resource
+
+
 class Api():
     """
     This module provides a high-level programmatic access to the HTTP API
@@ -284,17 +287,17 @@ class Api():
 
         return self._load_policy_file(policy_id, policy_file, HttpVerb.PATCH)
 
-    def rotate_other_api_key(self, resource_obj):
+    def rotate_other_api_key(self, resource: Resource):
         """
         This method is used to rotate a user/host's API key that is not the current user.
         To rotate API key of the current user use rotate_personal_api_key
         """
-        if resource_obj.resource_type not in ('user', 'host'):
+        if resource.type not in ('user', 'host'):
             raise Exception("Error: Invalid resource type")
 
         # Attach the resource type (user or host)
         query_params = {
-            'role': resource_obj.full_id()
+            'role': resource.full_id()
         }
         response = invoke_endpoint(HttpVerb.PUT, ConjurEndpoint.ROTATE_API_KEY,
                                    self._default_params,
--- a/conjur/cli.py
+++ b/conjur/cli.py
@@ -33,7 +33,7 @@ from conjur.login import LoginLogic, LoginController
 from conjur.logout import LogoutController, LogoutLogic
 from conjur.policy import PolicyData, PolicyLogic, PolicyController
 from conjur.variable import VariableLogic, VariableController, VariableData
-from conjur.user import UserController, UserResourceData, UserLogic
+from conjur.user import UserController, UserInputData, UserLogic
 from conjur.version import __version__
 
 # pylint: disable=too-many-statements
@@ -162,7 +162,7 @@ Copyright (c) 2020 CyberArk Software Ltd. All rights reserved.
                                                                                     '    conjur login -i admin \t\t\t'
                                                                                     'Prompts for password of the admin user to log in to Conjur server\n'
                                                                                     '    conjur login -i admin -p Myp@ssw0rd!\t'
-                                                                                    'Logs the admin user in to Conjur server and saves the user and password ' \
+                                                                                    'Logs the admin user in to Conjur server and saves the user and password '
                                                                                     'in the local cache (netrc file)'),
                                                          usage=argparse.SUPPRESS,
                                                          add_help=False,
@@ -186,7 +186,7 @@ Copyright (c) 2020 CyberArk Software Ltd. All rights reserved.
                                                           help='Log out from Conjur server and clear local cache',
                                                           description=self.command_description(logout_name, logout_usage),
                                                           epilog=self.command_epilog('conjur logout\t'
-                                                                                   'Logs out the user from the Conjur server and deletes the local ' \
+                                                                                   'Logs out the user from the Conjur server and deletes the local '
                                                                                    'cache (netrc file)'),
                                                           usage=argparse.SUPPRESS,
                                                           add_help=False,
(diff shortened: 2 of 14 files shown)


conjur/api.py
```diff
@@ -284,18 +284,17 @@ def update_policy_file(self, policy_id, policy_file):
 
         return self._load_policy_file(policy_id, policy_file, HttpVerb.PATCH)
 
-    def rotate_other_api_key(self, resource, resource_to_rotate):
+    def rotate_other_api_key(self, resource_obj):
```
abrahamko, line 287: does the `obj` suffix is needed?

conjur/cli.py
```diff
@@ -162,7 +162,7 @@ def run(self, *args):
                                                                                     '    conjur login -i admin \t\t\t'
```
abrahamko, line 162: won't multilie string serve you better than this infinite string concatenation?

conjur/host/host_controller.py
```diff
@@ -23,16 +24,18 @@ def __init__(self, client, host_resource_data):
...
-        new_api_key = self.client.rotate_other_api_key(resource,
-                                                          self.host_resource_data.host_to_update)
+
+        resource_obj = Resource(resource_type='host',
+                                resource_name=self.host_resource_data.host_to_update)
+        new_api_key = self.client.rotate_other_api_key(resource_obj)
         # pylint: disable=line-too-long
-        sys.stdout.write(f"Successfully rotated API key for '{self.host_resource_data.host_to_update}' " \
+        sys.stdout.write(f"Successfully rotated API key for '{self.host_resource_data.host_to_update}' "
                         f"New API key is: {new_api_key}\n")
```
abrahamko, line 39: add one more space here

conjur/resource.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    def __init__(self, **arg_params):
+        self.resource_type = arg_params['resource_type']
+        self.resource_name = arg_params['resource_name']
+
+    def full_id(self):
+        """
+        Method for building the full resource ID in the format 'user:someuser' for example
+        """
+        return ':'.join([self.resource_type, self.resource_name])
```
abrahamko, line 20: suggestion:
```suggestion
        return f"{self.resource_type}:{self.resource_name}")
```

conjur/resource.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+Resource module
+"""
+
+# pylint: disable=too-few-public-methods
+class Resource:
+    """
+    Used for representing Conjur resources
+    """
+    def __init__(self, **arg_params):
```
abrahamko, line 12: Why do you prefer it as dictionary instead of list of params?
```suggestion
    def __init__(self, type, name):
```

conjur/resource.py
```diff
@@ -0,0 +1,23 @@
...
+        self.resource_type = arg_params['resource_type']
```
abrahamko, line 13: `resource` prefix is redundant.
You already know it a resource.
so only `type` and `name` is enough, clear and concise.

conjur/user/user_logic.py
```diff
@@ -35,36 +35,38 @@ def rotate_api_key(self, resource_to_update):
...
+        # if the user provides their own user id and they are logged in,
+        # then for the sake of a successful outcome, we will ignore the input.
+        # Otherwise, the Conjur server will fail the request
+        if resource_to_update == current_logged_in_username:
+            resource_to_update = None
+        if resource_to_update is None:
             try:
-                logging.debug(f"Rotating API key for '{user_from_credential_store}'")
-                new_api_key = self.rotate_personal_api_key(user_from_credential_store,
+                logging.debug(f"Rotating API key for '{current_logged_in_username}'")
```
abrahamko, line 47: current and logged in both refer to the same user.
Choose one of them, and the other is redundant.
`current_username` or `logged_in_username`.

conjur/user/user_logic.py
```diff
@@ -35,36 +35,38 @@ def rotate_api_key(self, resource_to_update):
...
-            # if the user does not provide a user to rotate their API key,
-            # their own API key will be rotated
+        current_logged_in_username = loaded_credentials['login_id']
+
+        # if the user provides their own user id and they are logged in,
+        # then for the sake of a successful outcome, we will ignore the input.
+        # Otherwise, the Conjur server will fail the request
+        if resource_to_update == current_logged_in_username:
+            resource_to_update = None
+        if resource_to_update is None:
```
abrahamko, line 45: You can combine these checks:
if resource_to_update in (None, current_logged_in_username)

conjur/user/user_logic.py
```diff
@@ -35,36 +35,38 @@ def rotate_api_key(self, resource_to_update):
...
-                                   "use this command without any flags or options")
-        if resource_to_update is not None:
-            user_from_credential_store = resource_to_update
-            logging.debug(f"Rotating API key for '{user_from_credential_store}'")
-
-            new_api_key = self.rotate_other_api_key(user_from_credential_store)
-            logging.debug(f"Successfully rotated API key for '{user_from_credential_store}'")
-        else:
-            # if the user does not provide a user to rotate their API key,
-            # their own API key will be rotated
```
abrahamko, line 50: What happened to these comments? they are good!

conjur/user/user_logic.py
```diff
@@ -35,36 +35,38 @@ def rotate_api_key(self, resource_to_update):
         2. Rotate another users API key
         """
         loaded_credentials = self.extract_credentials_from_credential_store()
```
abrahamko, line 37: I think `credentials` is enough
```suggestion
        credentials = self.extract_credentials_from_credential_store()
```

abrahamko: you can prefix them with `current_` or `logged_in_` if you want, for clarity.

abrahamko (changes requested):
Good changes!
left some comments.

conjur/api.py
sigalsax: no but think it is clear that an object is needed and not a string for readability purposes but can change this to

`def rotate_other_api_key(self, resource: Resource):`

to achieve this readability

conjur/cli.py
sigalsax: as in `""""` instead?

conjur/host/host_controller.py
sigalsax: why? I have a space at the end of the previous line

conjur/resource.py
sigalsax: no reason. Updating

conjur/user/user_logic.py
sigalsax: Nice

conjur/resource.py
sigalsax: type is a reserved name in Python apparently

conjur/api.py
abrahamko: You may regret you proposed it...
Yes, this is wonderful! you should do it more across all the project!
You can do this for return value too, so take advantage of this too!
Also, static code analysis should identify this in case you use incorrect argument type.
Do we do this kind of analysis on our code? Will it find this mismatch in case wrong type was passed as arg?
In case it doesn't, you should make sure it does, so you could benefit from using this feature.

conjur/cli.py
abrahamko: As in
```python
'''
multiline string
'''
```

conjur/host/host_controller.py
abrahamko: to align the strings...

conjur/resource.py
abrahamko: so the param name can be resource_type, but the field can be `type`, so you could write `resource.type`

conjur/api.py
sigalsax: haha! Sounds like a separate PR. I want to keep this PR smaller but yes will implement!

conjur/cli.py
```diff
@@ -75,12 +75,16 @@ def main_epilog():
...
-    def command_epilog(*args):
+    def command_epilog(example, has_subcommand=False):
         """
         This method builds the footer for each command help screen.
         """
-        return '''Examples:
-    {}'''.format(*args)
+        refer_to_help = "Refer to the specific subcommand help for more details."
+        if has_subcommand:
+            return f'''Examples: \n{refer_to_help}
```
abrahamko, line 84: IS the newline required? Can you do the following?
```suggestion
            return f'''Examples:
            {refer_to_help}
```

conjur/resource.py
abrahamko: Better options suggested here: https://stackoverflow.com/questions/2417979/can-i-use-the-variable-name-type-as-function-argument-in-python

I would go with `typ` or `type_`, and only for arg name.
The field can and should stay `type`.

conjur/cli.py
sigalsax: it will turn into
```
       if has_subcommand:
            return f'''Examples:
{refer_to_help}
```
because of the sensitive formatting

conjur/host/host_controller.py
sigalsax: oh I thought you meant how the string will be printed to stdout. Sure on it!

conjur/user/user_logic.py
```diff
@@ -34,48 +34,50 @@ def rotate_api_key(self, resource_to_update):
...
-            # if the user does not provide a user to rotate their API key,
-            # their own API key will be rotated
+        logged_in_credentials = self.extract_credentials_from_credential_store()
+        logged_in_username = logged_in_credentials['login_id']
+
+        # if the user provides their own user id and they are logged in,
+        # then for the sake of a successful outcome, we will ignore the input.
+        # Otherwise, the Conjur server will fail the request.
+        # if the user does not provide a user to rotate their API key,
+        # their own API key will be rotated.
```
abrahamko, line 44: This comment reminds me of something @jonahx said to me once:
> I propose we call this a "Pinocchio" comment.
Just like Pinocchio was a puppet who wanted to be real boy, sometimes comments want to be a real variable name.

In your case, this comment want to be a real method...
Extract the inner code of this `if` to a method with proper name, and most of the comment will be redundant.

I think that the `else` deserve a method too 😉

conjur/resource.py
sigalsax: oh thought it was best practice for them to be called the same name so it would be `self.type = type`  or `self.res_type = res_type`

changing to `self.type = type_` and `self.name=name`

conjur/cli.py
sigalsax: then it makes it sensitive to formatting which will make it ugly

conjur/cli.py
abrahamko: still looks better IMO

sigalsax: updated

conjur/user/user_logic.py
```diff
@@ -21,11 +22,10 @@ class UserLogic:
...
     """
-    def __init__(self, conjurrc_data, credentials_from_store, client, resource):
+    def __init__(self, conjurrc_data, credentials_from_store, client):
         self.conjurrc_data=conjurrc_data
         self.credentials_from_store=credentials_from_store
         self.client=client
-        self.resource=resource
 
     # pylint: disable=logging-fstring-interpolation,line-too-long
     def rotate_api_key(self, resource_to_update):
```
abrahamko, line 31: Don't you have better name for that?
`username` maybe?

abrahamko (approved):
Last comment is a suggestion. do it if you feel it's right.

PR #154, opened 2021-01-28
Expand TLS version communication to 1.2 and 1.3

sigalsax:
When a user runs 'init' command, we open a socket with Conjur, fetch the certificate for them from the Conjur server and save it on their local machine.

We were previously limiting the context to to communicate over TLS 1.2 to fetch the certificate from the Conjur server. This change expands this to also support 1.3 and block anything less than 1.2. This option can now be configured via an environment variable.

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/conjur/ssl_service.py
+++ b/conjur/ssl_service.py
@@ -12,10 +12,11 @@ import socket
 from OpenSSL import SSL
 from OpenSSL.crypto import FILETYPE_PEM, dump_certificate
 
-_tls_methods = {
+_conjur_tls_methods = {
     "1.0": SSL.TLSv1_METHOD,
     "1.1": SSL.TLSv1_1_METHOD,
-    "1.2": SSL.TLSv1_2_METHOD
+    "1.2": SSL.TLSv1_2_METHOD,
+    "1.3": SSL.SSLv23_METHOD
 }
 
 # pylint: disable=too-few-public-methods
@@ -44,10 +45,20 @@ class SSLService:
         """
         Method for opening a socket to the Conjur server
         """
+        tls_version=os.getenv('CONJUR_TLS_VERSION')
         # Makes the TLS version configurable through an ENV variable
-        if os.getenv('TLS_VERSION') in _tls_methods:
-            ctx = SSL.Context(method=_tls_methods[os.getenv('TLS_VERSION')])
+        if tls_version in _conjur_tls_methods:
+            # pylint: disable=logging-fstring-interpolation
+            logging.debug("'CONJUR_TLS_VERSION' environment variable supplied. Establishing TLS "
+                          f"v{tls_version} connection...")
+            ctx = SSL.Context(method=_conjur_tls_methods[tls_version])
+            if tls_version == '1.3':
+                # disables all TLS versions expect for 1.3
+                cls.disable_tls_versions(ctx, support_1_3=True)
         else:
+            if tls_version:
+                logging.debug("Warning: 'CONJUR_TLS_VERSION' environment variable supplied but "
+                              "not in correct format (Example: CONJUR_TLS_VERSION=1.2)")
             # Despite its name, SSLv23_METHOD is the alias for PROTOCOL_TLS which allows for
             # negotiation between different TLS versions and chooses the highest supported server
             # protocol. See https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_SSLv23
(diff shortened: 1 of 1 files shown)


conjur/ssl_service.py
```diff
@@ -7,10 +7,17 @@
...
 import logging
+import os
 import socket
 from OpenSSL import SSL
 from OpenSSL.crypto import FILETYPE_PEM, dump_certificate
 
+_tls_methods = {
+    "1.0": SSL.TLSv1_METHOD,
+    "1.1": SSL.TLSv1_1_METHOD,
+    "1.2": SSL.TLSv1_2_METHOD
```
mbenita-Cyberark, line 18: I know TLSv1_3_METHOD is not exist but the user can still specify working with 1.3 only no? If I understand the SSL docs correctly

conjur/ssl_service.py
```diff
@@ -7,10 +7,17 @@
 """
 # Third party
 import logging
+import os
 import socket
 from OpenSSL import SSL
 from OpenSSL.crypto import FILETYPE_PEM, dump_certificate
 
+_tls_methods = {
```
mbenita-Cyberark, line 15: _tls_methods-> _TLS_METHODS
what do you think?
not sure what is our convention on this. but I assume this is a constant

conjur/ssl_service.py
```diff
@@ -37,7 +44,22 @@ def __connect(cls, hostname, port):
         """
         Method for opening a socket to the Conjur server
         """
-        ctx = SSL.Context(method=SSL.TLSv1_2_METHOD)
+        # Makes the TLS version configurable through an ENV variable
+        if os.getenv('TLS_VERSION') in _tls_methods:
+            ctx = SSL.Context(method=_tls_methods[os.getenv('TLS_VERSION')])
```
mbenita-Cyberark, line 49: who is encharge of this env variable?  
1) recommend `tls_version = os.getenv('TLS_VERSION')` instead of reading it twice. I don't know how paranoid should I be but theoretically it could be change  after the if by an attacker (was it too paranoid?)
2) if we are not maintaining this env variable how can we be sure it is in the right format? maybe try catch incase there is a value but it is not in the right format, or ssl?

mbenita-Cyberark (commented):
Don't have much knowledge about the library. so I hope this helps

conjur/ssl_service.py
```diff
@@ -37,7 +44,22 @@ def __connect(cls, hostname, port):
...
+            # negotiation between different TLS versions and chooses the highest supported server
+            # protocol. See https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_SSLv23
+            # specifically PROTOCOL_SSLv23 and PROTOCOL_TLS
+            ctx = SSL.Context(method=SSL.SSLv23_METHOD)
+            # Using SSLv23 by default enables all types of negotiation and we want to limit
+            # negotiation to only 1.2 and 1.3 versions
+            ctx.set_options(SSL.OP_NO_SSLv2)
+            ctx.set_options(SSL.OP_NO_SSLv3)
+            ctx.set_options(SSL.OP_NO_TLSv1)
+            ctx.set_options(SSL.OP_NO_TLSv1_1)
```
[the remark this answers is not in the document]
sigalsax: yes it will use 1.3 but we don't want to block 1.2 because if 1.3 is not available, then there will be a failure. The client provides a list of versions (1.2,1.3) and the server will pick the highest version (1.2 for example)

conjur/ssl_service.py
sigalsax: correct TLSv1_3_METHOD doesnt exist but SSLv23_METHOD exists which allows negotiation with 1.3. We can say that if the user specifies 1.3, then 1.2 will also be blocked (`ctx.set_options(SSL.OP_NO_TLSv1_2)`

Something like 
if os.getenv('TLS_VERSION') == "1.3":
  ```
          ctx.set_options(SSL.OP_NO_SSLv2)
          ctx.set_options(SSL.OP_NO_SSLv3)
          ctx.set_options(SSL.OP_NO_TLSv1)
          ctx.set_options(SSL.OP_NO_TLSv1_1)
          ctx.set_options(SSL.OP_NO_TLSv1_2)
  ```

conjur/ssl_service.py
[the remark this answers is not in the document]
mbenita-Cyberark: Again I'm new to that stuff so maybe I'm wrong. An attacker will want to downgrade protocol security. that is why the server should allow small number of protocols options as possible. 
This is up to us to decide what should be the default (1.3 only or 1.3 and 1.2) I guess we should choose the nginx tls default version

conjur/ssl_service.py
sigalsax: tried to be uniform with the pyopenssl library I was using

Screen Shot 2021-01-28 at 3 29 19 PM

conjur/ssl_service.py
sigalsax: 1. updated to be variable for readability
2. not sure what you mean. it is just an environment var that can be set like `export TLS_VERSION=1.1`. What do you mean proper format?

conjur/ssl_service.py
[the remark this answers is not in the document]
mbenita-Cyberark: talk on the phone, keeping 1.2 1.3

mbenita-Cyberark (approved):

PR #171, opened 2021-02-04
Fix title rendering

sigalsax:
### What does this PR do?
Makes minor changes as per UX changes demonstrated here in the [UX guide ](https://ljfz3b.axshare.com/#id=x8ktq8&p=conjur_help__init&g=1)

### DIFF
--- a/conjur/cli.py
+++ b/conjur/cli.py
@@ -81,7 +81,7 @@ To start using Conjur with your environment, you must first initialize the confi
         """
         This method builds the footer for each command help screen.
         """
-        refer_to_help = "See more details in each subcommand’s help:"
+        refer_to_help = "See more details in each subcommand's help:"
         if subcommands:
             res = ""
             for subcommand in subcommands:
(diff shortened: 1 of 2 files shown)


tzheleznyak (dismissed):

conjur/cli.py
```diff
@@ -81,7 +81,7 @@ def command_epilog(example, command=None, subcommands=None):
         """
         This method builds the footer for each command help screen.
         """
-        refer_to_help = "See more details in each subcommand’s help:"
+        refer_to_help = "See more details in each subcommand\'s help:"
```
orenbm, line 84: the backslash is not a mistake?

orenbm (dismissed):

conjur/cli.py
sigalsax: I added it in because I thought it would error but I see now it doesnt and will only error when I use `''` over `""`

tzheleznyak (approved):

PR #190, opened 2021-02-11
Add ability to run init in insecure mode

sigalsax:
### What does this PR do?
The purpose of this PR is to enhance the UX for the user when running in insecure mode. Previously the user would have to physically create the conjurrc file if they wanted to run the init in insecure mode b/c the command didnt support this flag. This addition allows the user to run init in insecure mode and get a blank value for the cert_file entry, creating the following conjurrc

```
---
account: cucumber
appliance_url: https://ec2-3-126-247-26.eu-central-1.compute.amazonaws.com
cert_file: ''
plugins: []
```
Resolves -

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/conjur/api/client.py
+++ b/conjur/api/client.py
@@ -149,7 +149,7 @@ class Client():
     # Technical debt: refactor when time permits because this function
     # doesn't belong here
     @staticmethod
-    def initialize(url, account, cert, force):
+    def initialize(url, account, cert, force, ssl_verify):
         """
         Initializes the client, creating the .conjurrc file
         """
@@ -163,7 +163,8 @@ class Client():
 
         input_controller = InitController(conjurrc_data,
                                           init_logic,
-                                          force)
+                                          force,
+                                          ssl_verify)
         input_controller.load()
 
     ### API passthrough
--- a/conjur/cli.py
+++ b/conjur/cli.py
@@ -564,11 +564,12 @@ Copyright (c) 2021 CyberArk Software Ltd. All rights reserved.
             sys.exit(0)
 
     @classmethod
-    def handle_init_logic(cls, url=None, name=None, certificate=None, force=None):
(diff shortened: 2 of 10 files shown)


tzheleznyak:
The change is user facing and should be documented in changelog, not?

conjur/init/init_controller.py
```diff
@@ -36,11 +41,13 @@ def load(self):
         """
         Method that facilitates all method calls in this class
         """
-        fetched_certificate = self.get_server_certificate()
-        self.write_certificate(fetched_certificate)
+        if self.ssl_verify is True:
+            fetched_certificate = self.get_server_certificate()
+            self.write_certificate(fetched_certificate)
+        else:
+            self.conjurrc_data.cert_file=""
```
tzheleznyak, line 48: why we need to do this?

sigalsax: because it is set to None at the argparse level because the user did not provide the certificate so we want to purposefully set it to "" so that the entry is still printed because we have in init_logic.py 
```
            # Ensures that there are no None fields written to conjurrc
            for attr,value in conjurrc_data.__dict__.items():
                if value is not None:
                    _pretty_print_object[str(attr)]=value
```
to ensure that the cert file is still being printed but now that i think about it I can remove this check because all values are filled by our work in the CLI

sigalsax:
> The change is user facing and should be documented in changelog, not?

Good question but no insecure is a flag part of all commands, not just init and we didnt go into implementation details in the CHANGELOG. We will document but not in changelog

conjur/init/init_controller.py
tzheleznyak: Cant we set it to "" in argparse level?

sigalsax: ive updated it always

tzheleznyak (approved):

PR #195, opened 2021-02-14
Fixes the bug where an empty cert_file entry allows for a user to login

sigalsax:
### What does this PR do?
When a user runs in insecure mode, the conjurrc looked like:

```
---
account: cucumber
appliance_url: https://conjur-server
cert_file: ''
plugins: []
```
When they send a follow-up login command, the empty cert_file was being translated to FALSE. We don't want to make this implicit conclusion and only run in insecure mode if the user declares to do so

### DIFF
--- a/conjur/logic/login_logic.py
+++ b/conjur/logic/login_logic.py
@@ -38,8 +38,13 @@ class LoginLogic:
 
         if ssl_verify is False:
             certificate_path = False
-        elif credential_data.machine.startswith("https"):
-            certificate_path = conjurrc.cert_file
+        elif ssl_verify and credential_data.machine.startswith("https"):
+            # Catches the case where a user does not run in insecure mode but the
+            # .conjurrc cert_file entry is empty
+            if conjurrc.cert_file == '':
+                certificate_path = True
+            else:
+                certificate_path = conjurrc.cert_file
 
         # pylint: disable=logging-fstring-interpolation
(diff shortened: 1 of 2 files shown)


conjur/logic/login_logic.py
```diff
@@ -38,8 +38,13 @@ def get_api_key(cls, ssl_verify, credential_data, password, conjurrc):
 
         if ssl_verify is False:
             certificate_path = False
-        elif credential_data.machine.startswith("https"):
-            certificate_path = conjurrc.cert_file
+        elif ssl_verify is True and credential_data.machine.startswith("https"):
```
tzheleznyak, line 41:
```suggestion
        elif ssl_verify and credential_data.machine.startswith("https"):
```

conjur/logic/login_logic.py
```diff
@@ -38,8 +38,13 @@ def get_api_key(cls, ssl_verify, credential_data, password, conjurrc):
...
+            # Catches the case where a user does not run in insecure mode but the
+            # .conjurrc cert_file entry is empty
+            if conjurrc.cert_file == '':
+                certificate_path = True
```
tzheleznyak, line 45: It seems certificate_path takes values from diffrent types boolean and the type of conjurrc.cert_file.

We need to find a way to be type consistent for each variable

sigalsax: as explained requests takes boolean and string

tzheleznyak (approved):

PR #196, opened 2021-02-15
Add support for CA signed certificates

### LINKED ISSUE #178: SSL: CERTIFICATE_VERIFY_FAILED when running login
When using the CLI in a Centos environment, we are successful in running `init`, saving the pem to the machine received from the Conjur server and generating the conjurrc. Upon sending a follow-up request (login), the request is failing on CERTIFICATE_VERIFY_FAILED error (see image below)

Returned error:

### Known:
1. Ruby CLI works, Python CLI doesn't
2. Both returned certificates from both the flows are the same

### Troubleshooting:
1. Ran `strace` and found that successful and non-successful flows of reading the certificate are the same
Successful scenario:
Screen Shot 2021-02-10 at 3 24 52 PM
Not successful scenario
Screen Shot 2021-02-10 at 3 25 35 PM

2. As a sanity check, tried to switch out the Python certificate in `.conjurrc` for Ruby one and still failure persists.

3. When running `openssl s_client -connect google.com:443 -tls1_2`

Screen Shot 2021-02-11 at 1 26 09 PM

Screen Shot 2021-02-11 at 1 24 45 PM

4. As a sanity check, I ran the Python CLI against a Layer 4 LB and was successful
5. Ran with `--insecure` flag to ignore the ssl verification and commands run successfully
### Hunch:
Possible proxy issues

sigalsax:
### What does this PR do?
When using the CLI in an environment with signed certificates, we are successful in running init, saving the pem to the machine received from the Conjur server and generating the conjurrc. Upon sending a follow-up request (login), the request is failing on CERTIFICATE_VERIFY_FAILED error. This fix adds the signed-certificate flow

At this time we will not add the CA cert flow to our testing infrastructure. There is a [separate ticket for this](https://github.com/cyberark/conjur-api-python3/issues/200)
Resolves #178 

#### Change log

#### Test coverage

#### Documentation
Will be added later

### DIFF
--- a/conjur/api/client.py
+++ b/conjur/api/client.py
@@ -13,7 +13,7 @@ import logging
 # Internals
 import netrc
 
-from conjur.errors import OperationNotCompletedException
+from conjur.errors import InvalidOperationException
 from conjur.util import util_functions
 from conjur.api import Api
 from conjur.config import Config as ApiConfig
@@ -90,16 +90,16 @@ class Client():
                 # Raise exception if the client was initialized in insecure mode
                 # but a follow-up request is run without the insecure flag
                 if ssl_verify is True and loaded_config['ca_bundle'] == '':
-                    raise OperationNotCompletedException(cause="The client was initialized without certificate verification, "
-                                        "even though the command was ran with certificate verification enabled. ",
+                    raise InvalidOperationException(cause="The client was initialized without certificate verification, "
+                                        "even though the command was ran with certificate verification enabled.",
                                         solution="To continue communicating with the server insecurely, run the command "
-                                        "again with the --insecure flag. Otherwise, reinitialize the client.")
+                                        "again with the --insecure flag. Otherwise, reinitialize the client`")
 
                 logging.debug("Fetched connection details: "
                               f"{{'account': {loaded_config['account']}, "
                               f"'appliance_url': {loaded_config['url']}, "
                               f"'cert_file': {loaded_config['ca_bundle']}}}")
-            except OperationNotCompletedException:
+            except InvalidOperationException:
                 raise
             # TODO add error handling for when conjurrc field doesn't exist
             except Exception as exc:
--- a/conjur/api/client.py
+++ b/conjur/api/client.py
@@ -51,7 +51,7 @@ class Client():
 
     # The method signature is long but we want to explicitly control
     # what parameters are allowed
-    # pylint: disable=too-many-arguments,too-many-locals
+    # pylint: disable=too-many-arguments,too-many-locals,too-many-branches
     def __init__(self,
                  account=None,
                  api_key=None,
@@ -87,12 +87,11 @@ class Client():
                     if field_value:
                         on_disk_config[field_name] = field_value
                 loaded_config = on_disk_config
-
                 # Raise exception if the client was initialized in insecure mode
                 # but a follow-up request is run without the insecure flag
                 if ssl_verify is True and loaded_config['ca_bundle'] == '':
                     raise OperationNotCompletedException(cause="The client was initialized without certificate verification, "
-                                        "even though the command was ran with certificate verification turned on. ",
+                                        "even though the command was ran with certificate verification enabled. ",
                                         solution="To continue communicating with the server insecurely, run the command "
                                         "again with the --insecure flag. Otherwise, reinitialize the client.")
 
@@ -100,7 +99,8 @@ class Client():
                               f"{{'account': {loaded_config['account']}, "
                               f"'appliance_url': {loaded_config['url']}, "
                               f"'cert_file': {loaded_config['ca_bundle']}}}")
-
+            except OperationNotCompletedException:
+                raise
             # TODO add error handling for when conjurrc field doesn't exist
             except Exception as exc:
                 raise ConfigException(exc) from exc
--- a/conjur/api/client.py
+++ b/conjur/api/client.py
@@ -13,15 +13,12 @@ import logging
 # Internals
 import netrc
 
+from conjur.errors import OperationNotCompletedException
 from conjur.util import util_functions
 from conjur.api import Api
-from conjur.api.ssl_client import SSLClient
 from conjur.config import Config as ApiConfig
 from conjur.constants import DEFAULT_NETRC_FILE
-from conjur.controller import InitController
-from conjur.logic import InitLogic
 from conjur.util import CredentialsFromFile
-from conjur.data_object import ConjurrcData
 from conjur.resource import Resource
 
 
@@ -39,7 +36,7 @@ class ConfigException(Exception):
     SDK in their code.
     """
 
-
+# pylint: disable=logging-fstring-interpolation,line-too-long
 class Client():
     """
     Client
@@ -90,7 +87,15 @@ class Client():
                     if field_value:
                         on_disk_config[field_name] = field_value
                 loaded_config = on_disk_config
-                # pylint: disable=logging-fstring-interpolation
+
+                # Raise exception if the client was initialized in insecure mode
+                # but a follow-up request is run without the insecure flag
+                if ssl_verify is True and loaded_config['ca_bundle'] == '':
+                    raise OperationNotCompletedException(cause="The client was initialized without certificate verification, "
+                                        "even though the command was ran with certificate verification turned on. ",
+                                        solution="To continue communicating with the server insecurely, run the command "
+                                        "again with the --insecure flag. Otherwise, reinitialize the client.")
+
                 logging.debug("Fetched connection details: "
                               f"{{'account': {loaded_config['account']}, "
                               f"'appliance_url': {loaded_config['url']}, "
@@ -145,27 +150,6 @@ class Client():
             logging.basicConfig(level=logging.DEBUG, format=self.LOGGING_FORMAT)
         else:
             logging.basicConfig(level=logging.WARN, format=self.LOGGING_FORMAT)
-
-    # Technical debt: refactor when time permits because this function
-    # doesn't belong here
-    @staticmethod
-    def initialize(url, account, cert, force, ssl_verify):
-        """
-        Initializes the client, creating the .conjurrc file
-        """
-        ssl_service = SSLClient()
-
-        conjurrc_data = ConjurrcData(url,
-                                     account,
-                                     cert)
-
-        init_logic = InitLogic(ssl_service)
-
-        input_controller = InitController(conjurrc_data,
-                                          init_logic,
-                                          force,
-                                          ssl_verify)
-        input_controller.load()
 
     ### API passthrough
 
--- a/conjur/errors.py
+++ b/conjur/errors.py
@@ -20,8 +20,8 @@ class OperationNotCompletedException(Exception):
     Exception for when an operation was not completed successfully
     and CLI is left in instable state
     """
-    def __init__(self, message="Error: Failed to run command to completion."):
-        self.message = message
+    def __init__(self, cause="", solution=""):
+        self.message = "Error: Failed to run command to completion. " + cause + solution
         super().__init__(self.message)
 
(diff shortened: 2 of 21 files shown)


test/test_unit_client.py
```diff
@@ -129,19 +128,6 @@ def test_client_can_raise_netrc_exception_error(self, mock_cred, mock_api_instan
         with self.assertRaises(Exception):
             Client(url='https://myurl', account='myacct', login_id='mylogin', ca_bundle="mybundle", ssl_verify=False)
 
-    @patch('conjur.api.Api')
```
sigalsax, line 132: removed because it was duplicated code

test/util/test_path_provider.py
```diff
@@ -85,8 +88,3 @@ def certificate_path(self):
 
     def get_policy_path(self, type: str):
         return os.path.join(self.HELPERS_FILES_DIR, "test_config", f"{type}_policy.yml")
-
-    @property
```
sigalsax, line 89: removed because it was duplicated

conjur/errors.py
```diff
@@ -20,8 +20,8 @@ class OperationNotCompletedException(Exception):
     Exception for when an operation was not completed successfully
     and CLI is left in instable state
     """
-    def __init__(self, message="Error: Failed to run command to completion."):
-        self.message = message
+    def __init__(self, cause="", solution=""):
+        self.message = "Error: Failed to run command to completion. " + cause + solution
```
tzheleznyak, line 24: fstring it

conjur/wrapper/http_wrapper.py
```diff
@@ -68,8 +74,8 @@ def invoke_endpoint(http_verb, endpoint, params, *args, check_errors=True,
             response.raise_for_status()
         except requests.exceptions.HTTPError as http_error:
             if response.text:
-                logging.debug(requests.exceptions.HTTPError(f'{http_error.response.status_code}' \
-                                                            f' {http_error.response.reason}' \
+                logging.debug(requests.exceptions.HTTPError(f'{http_error.response.status_code}'
```
tzheleznyak, line 77: Just noticed that in some of your logs you do '' and in some you do ""

Lets try to be consistent with this

test/test_unit_client.py
```diff
@@ -81,13 +79,13 @@ def test_client_passes_init_parameters_with_none_provided(self):
...
         client.initialize.assert_called_once_with(None, None, None, False)
 
-    @patch('conjur.api.client.ApiConfig', new=MissingMockApiConfig)
-    def test_client_init_initialize_calls_load_properly(self):
-        client = Client
-        mock_init_controller = InitController
-        mock_init_controller.load = MagicMock()
-        Client.initialize('someurl', 'someaccount', '/some/path/to/pem', False, True)
-        mock_init_controller.load.assert_called_once()
+    # @patch('conjur.api.client.ApiConfig', new=MissingMockApiConfig)
```
tzheleznyak, line 82: why this commented code

test/test_unit_client.py
```diff
@@ -291,19 +289,6 @@ def test_client_increases_logging_with_debug_flag(self, mock_logging, mock_creds
 
         mock_logging.assert_called_once_with(format=Client.LOGGING_FORMAT, level=logging.DEBUG)
 
-    @patch('conjur.api.client.Api')
-    def test_client_passes_default_account_to_api_initializer_if_none_is_provided(self, mock_api_instance):
-        Client(url='http://myurl', login_id='mylogin', password='mypass',
```
tzheleznyak, line 296: why this test is removed?

test/test_unit_client.py
sigalsax: because the UX wasnt finalized so didnt want to write UTs. Updating now

conjur/api/client.py
```diff
@@ -90,12 +87,20 @@ def __init__(self,
...
                 loaded_config = on_disk_config
-                # pylint: disable=logging-fstring-interpolation
+                # Raise exception if the client was initialized in insecure mode
+                # but a follow-up request is run without the insecure flag
+                if ssl_verify is True and loaded_config['ca_bundle'] == '':
+                    raise OperationNotCompletedException(cause="The client was initialized without certificate verification, "
+                                        "even though the command was ran with certificate verification enabled. ",
+                                        solution="To continue communicating with the server insecurely, run the command "
+                                        "again with the --insecure flag. Otherwise, reinitialize the client.")
+
```
mbenita-Cyberark, line 97: not sure this is OperationNotCompletedException. more like invalidOperationException, or MissingArgumentsException

conjur/logic/login_logic.py
```diff
@@ -41,10 +43,13 @@ def get_api_key(cls, ssl_verify, credential_data, password, conjurrc):
...
-            if conjurrc.cert_file == '':
-                certificate_path = True
-            else:
-                certificate_path = conjurrc.cert_file
+            if conjurrc.cert_file == '' and ssl_verify:
+                raise OperationNotCompletedException(cause="The client was initialized without certificate verification, "
+                                    "even though the command was ran with certificate verification enabled. ",
+                                    solution="To continue communicating with the server insecurely, run the command "
+                                    "again with the --insecure flag. Otherwise, reinitialize the client.")
+
```
mbenita-Cyberark, line 51: not sure you want to raise the ui message here. and if so, why in the client.py as well?
I would change OperationNotCompletedException to something else.
For my opinion the OperationNotCompletedException should be used in one of two.
1) the exception caused by unfinished operation
2) you want to indicate to the user that this operation is not completed (but actually all exceptions indicate this)

conjur/api/client.py
```diff
@@ -90,12 +87,20 @@ def __init__(self,
                     if field_value:
                         on_disk_config[field_name] = field_value
                 loaded_config = on_disk_config
-                # pylint: disable=logging-fstring-interpolation
+                # Raise exception if the client was initialized in insecure mode
+                # but a follow-up request is run without the insecure flag
+                if ssl_verify is True and loaded_config['ca_bundle'] == '':
```
oburstein-hub, line 92: when ssl_verify is true ?

conjur/wrapper/http_wrapper.py
```diff
@@ -55,11 +53,19 @@ def invoke_endpoint(http_verb, endpoint, params, *args, check_errors=True,
...
-                              verify=ssl_verify,
-                              auth=auth,
-                              headers=headers)
+    try:
+        response = request_method(url, *args,
+                                  params=query,
+                                  verify=True,
+                                  auth=auth,
+                                  headers=headers)
+    except requests.exceptions.SSLError:
```
mbenita-Cyberark, line 62: this is the exception for the wrong cert I guess.. are there anymore exceptions that could be caused by wrong cert?

conjur/logic/login_logic.py
```diff
@@ -41,10 +43,13 @@ def get_api_key(cls, ssl_verify, credential_data, password, conjurrc):
         elif ssl_verify and credential_data.machine.startswith("https"):
             # Catches the case where a user does not run in insecure mode but the
             # .conjurrc cert_file entry is empty
-            if conjurrc.cert_file == '':
-                certificate_path = True
-            else:
-                certificate_path = conjurrc.cert_file
+            if conjurrc.cert_file == '' and ssl_verify:
```
oburstein-hub, line 46: how is this section related to support of CA signed certificates ?

conjur/wrapper/http_wrapper.py
```diff
@@ -55,11 +53,19 @@ def invoke_endpoint(http_verb, endpoint, params, *args, check_errors=True,
...
+        response = request_method(url, *args,
```
oburstein-hub, line 63: is this a fallback to self-signed certificate ? If yes - please add a comment in code for it

oburstein-hub (commented):
If I understood that the logic of fallback to self-signed is in the place I have mentioned - then it looks OK

conjur/api/client.py
sigalsax: ssl_verify is the mode the user runs the CLI (T/F). They can run the CLI in --insecure mode then ssl_verify = False. Otherwise, it will be True

conjur/logic/login_logic.py
sigalsax: if upon initialization of the CLI, the user runs `conjur --insecure init -u https://conjur-server` then
 ssl_verify=False and the cert_file param in the conjurrc will be set to ''

```
---
account: cucumber
appliance_url: https://conjur-server.pem
cert_file: ''
plugins: []
```
So this check is to determine if the cli was initialized in insecure mode but the follow-up command (login) was run without `--insecure`

conjur/wrapper/http_wrapper.py
sigalsax: updated this spot. The rest are in testing files I believe so I prefer not to touch them in this PR because it will get messy and cloud this PR

conjur/logic/login_logic.py
sigalsax: @mbenita-Cyberark where would you catch it? I can raise the error and keep raising it through the many layers but where would we catch it and print this specific error. 

My thoughts were the following:
1. OperationNotCompletedException (now changed to `InvalidOperationException`) is a very general error so I didn't want to catch it in the cli.py because I have that UI message I want to print.
2. I also didn't want to catch it in everyone of the controllers because that is not scalable
3. This error is also appropriate to the SDK whose entrypoint is client.py so I want this message to also be seen by those using the SDK.
4. 
What do you think?

conjur/wrapper/http_wrapper.py
sigalsax: if self-signed cert and verify=True the following error will be raised:
`'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain`

conjur/logic/login_logic.py
mbenita-Cyberark: 1. 100%
2,3. why not here? I think the controller is in charge of the UI interaction and therefore it should raise the exception with the UI message.
correct me if I'm wrong the sdk can not use the --insecure flag the way it is written in the error message.
The error for the sdk can be thrown here (with some message changes) as it is now.
4.

sigalsax: they can through the SDK. See `ssl_verify`. when conjurrc's cert_file='' and the user feeds ssl_verify=False then the error should be raised (with that message)
See init of client
```
    def __init__(self,
                 account=None,
                 api_key=None,
                 ca_bundle=None,
                 debug=False,
                 http_debug=False,
                 login_id=None,
                 password=None,
                 ssl_verify=True,
                 url=None):
```

sigalsax: > I think the controller is in charge of the UI interaction and therefore it should raise the exception with the UI message.

Agree and didnt feel right putting it here

conjur/errors.py
```diff
@@ -20,8 +20,16 @@ class OperationNotCompletedException(Exception):
...
+    def __init__(self, cause="", solution=""):
+        self.message = f"Error: Failed to run command to completion. Reason: {cause} {solution}"
+        super().__init__(self.message)
+
+class InvalidOperationException(Exception):
+    """
+    Exception for when the operation performed was invalid
+    """
+    def __init__(self, cause="", solution=""):
+        self.message = f"Error: Invalid operation. Reason: {cause} {solution}"
```
tzheleznyak, line 32:
```suggestion
        self.message = f"Error: Invalid operation. Reason: {cause} Solution: {solution}"
```

tzheleznyak (approved):
Looks good. Left small comment about one of the log messages

sigalsax:
@eladkug opened a ticket to add CA signed certs to our testing

tzheleznyak (approved):
LGTM

sigalsax:
approval from @moticless (:

PR #220, opened 2021-02-28
Update change-password tests to change non-admin passwords and fix failing UTs

sigalsax:
### What does this PR do?
Password complexity restriction was a feature introduced in a relatively newer version of Conjur. Before release we need to check against these older server versions and therefore when running the integration tests for change-password, the tests were failing because an error was not being raised and the actual password was being changed which leads to failing tests on future runs. This fixes this by attempting to change the password on a dummy user instead of admin's.

This PR also gives our broken UT build. The instability was coming the way at which the Jenkins pipeline is running the tests. 

Resolves -

#### Change log

#### Test coverage

#### Documentation


tzheleznyak (approved):
LGTM

PR #222, opened 2021-03-02
Update login error handling

### LINKED ISSUE #223: 502 bad gateway error show misleading instruction
## Is your feature request related to a problem? Please describe.

When getting http error in login command we get "Check your credential" instruction

## Describe the solution you would like

Every http error should show it dedicated instruction. and if unhandled, don't show instruction

## Additional context
The error shown
image

sigalsax:
### What does this PR do?

Previously, all errors were returning 'Check your credentials' solution response regardless of the type of error being returned. This commit fixes this by differentiating between the status codes to return the relevant error message.

Resolves #223

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/conjur/cli.py
+++ b/conjur/cli.py
@@ -22,6 +22,7 @@ import requests
 from conjur.api import SSLClient
 from conjur.errors import CertificateVerificationException
 from conjur.errors_messages import INCONSISTENT_VERIFY_MODE_MESSAGE
+from conjur.util.util_functions import determine_status_code_specific_error_messages
 from conjur.wrapper import ArgparseWrapper
 from conjur.api.client import Client
 from conjur.constants import DEFAULT_NETRC_FILE, DEFAULT_CONFIG_FILE
@@ -554,9 +555,8 @@ Copyright (c) 2021 CyberArk Software Ltd. All rights reserved.
         except requests.exceptions.HTTPError as server_error:
             logging.debug(traceback.format_exc())
             # pylint: disable=no-member
-            if hasattr(server_error.response, 'status_code') and str(server_error.response.status_code) == '401':
-                sys.stdout.write("Failed to log in to Conjur. Unable to authenticate with Conjur. "
-                    f"Reason: {server_error}. Check your credentials and try again.\n")
+            if hasattr(server_error.response, 'status_code'):
+                sys.stdout.write(determine_status_code_specific_error_messages(server_error))
             else:
                 sys.stdout.write(f"Failed to execute command. Reason: {server_error}\n")
 
--- a/conjur/cli.py
+++ b/conjur/cli.py
@@ -551,9 +551,15 @@ Copyright (c) 2021 CyberArk Software Ltd. All rights reserved.
             logging.debug(traceback.format_exc())
             sys.stdout.write(f"Error: No such file or directory: '{not_found_error.filename}'\n")
             sys.exit(1)
-        except requests.exceptions.HTTPError as client_error:
+        except requests.exceptions.HTTPError as server_error:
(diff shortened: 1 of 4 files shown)


conjur/cli.py
```diff
@@ -551,9 +551,15 @@ def run(self, *args):
             logging.debug(traceback.format_exc())
             sys.stdout.write(f"Error: No such file or directory: '{not_found_error.filename}'\n")
             sys.exit(1)
-        except requests.exceptions.HTTPError as client_error:
+        except requests.exceptions.HTTPError as server_error:
```
mbenita-Cyberark, line 554: 100% agree with the renaming

tzheleznyak (dismissed):
When another http error code need to be handled here it need to be moved to another file

conjur/cli.py
```diff
@@ -551,9 +551,15 @@ def run(self, *args):
...
             logging.debug(traceback.format_exc())
-            sys.stdout.write(f"Failed to execute command. Reason: {client_error}\n")
+            # pylint: disable=no-member
+            if hasattr(server_error.response, 'status_code') and str(server_error.response.status_code) == '401':
+                sys.stdout.write("Failed to log in to Conjur. Unable to authenticate with Conjur. "
+                    f"Reason: {server_error}. Check your credentials and try again.\n")
```
mbenita-Cyberark, line 559: what we do here is forwording the http error message with some cli instructions 
I think not only 401 error is relevant but more (almost all 500's could be followed with some general instructions) so I think we should extract the handle to outer function where we could add handeling for other errors over time.
BTW CLI can be very sophisticate and for every error it could narrow down the actual problem and give the user very good and informative instructions 

```suggestion
            if hasattr(server_error.response, 'status_code') #This should check if we can filter by error code
                print_user_informative_error(server_error)
        def print_user_informative_error(server_error):
            if str(server_error.response.status_code) == '401':
                   sys.stdout.write("Failed to log in to Conjur. Unable to authenticate with Conjur. "
                       f"Reason: {server_error}. Check your credentials and try again.\n")
```

conjur/controller/login_controller.py
```diff
@@ -93,13 +92,7 @@ def get_api_key(self, conjurrc):
         """
         Method to fetch the user/host's API key from Conjur which is to be added to the netrc
         """
-        try:
-            self.credential_data.api_key = self.login_logic.get_api_key(self.ssl_verify,
-                                                                        self.credential_data,
-                                                                        self.user_password,
-                                                                        conjurrc)
-        except CertificateVerificationException:
-            raise
```
mbenita-Cyberark, line 102: why removing this?

sigalsax: because it is caught in the `except Exception as server_error` of the `cli.py` so we don't need this

conjur/cli.py
sigalsax: "BTW CLI can be very sophisticate and for every error it could narrow down the actual problem and give the user very good and informative instructions"
how do you suggest?

sigalsax:
@tzheleznyak implemented your suggestion now (:

conjur/controller/login_controller.py
mbenita-Cyberark: Now see this is just catch and throw the same so really no needed

conjur/cli.py
mbenita-Cyberark: for some http error we can give a set of instruction.
for example if it is a 502 we can tell the user to check if conjur is up. if we want to be even more sophisticate we can check for server /info and if it is a standby we can raise "note that you are trying to access a standby machine"

mbenita-Cyberark (approved):

PR #224, opened 2021-03-02
Add release automation doc

sigalsax:
### What does this PR do?
This PR adds the release automation design doc to the project

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/design/release_automation.md
+++ b/design/release_automation.md
@@ -1,0 +1,87 @@
+# Release Automation
+
+This document is the design document for automating our release process for the Python CLI/SDK project. We have a requirement to support four different platforms (RHEL 7/8, macOS, and Windows) and need to create automated flows for each of them. 
+
+#### Milestone 1
+
+1. Spin up 4 machines/containers, pack the CLI executable, and add to artifacts of the build
+   -  The artifacts will be used for testing
+
+#### Milestone 2
+
+1. Run the integration tests on those machines as part of our pipeline. We want to run the tests on new machines that don't have Python or any other dependencies to ensure our tests accurately represent user environments.
+2. Run integration tests against older Conjur servers (Conjur enterprise v5.6.3 and Conjur OSS v1.2.0).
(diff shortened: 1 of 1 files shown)


design/release_automation.md
```diff
@@ -0,0 +1,87 @@
...
+This document is the design document for automating our release process for the Python CLI/SDK project. We have a requirement to support four different platforms (RHEL 7/8, macOS, and Windows) and need to create automated flows for each of them. 
+
+#### Milestone 1
+
+1. Spin up 4 machines/containers, pack the CLI executable, and add to artifacts of the build
+   -  The artifacts will be used for testing
+
+#### Milestone 2
+
+1. Run the integration tests on those machines as part of our pipeline. We want to run the tests on new machines that don't have Python or any other dependencies to ensure our tests accurately represent user environments.
```
hughsaunders, line 12: Very common for linux systems to have an out of date 'system' python. May be worth testing this case also.

garymoon: This, to the extent that LTS distros will still be running Python 2 (RHEL will be the biggest offender here) and any LTS distro will be eventually running EoL Python well before it goes EoL itself.

ismarc:
With the breadth of platforms needing to be validated, I would suggest one of two approaches:
1. Validate within docker containers that map to the distributions being verified (for Linux).  I believe this validation would be representative of what a bare VM would look like and allows for faster (and thus a broader potential set to test against) testing.
2. Spin up servers to test against rather than directly using Jenkins agents.  The Jenkins agents have a necessary baseline of software, including python3, that may cause the tests to not be representative.  If the broad set of platform tests isn't happening for every build, but instead prior to release with a limit-platform test occurring for each PR/merge, then the time to spin up the servers and verify shouldn't have a large impact on end-users.

MacOS and Windows is more complicated, and testing against Jenkins agents there will greatly simplify the process and have less chances for agent-related conflicts.

sigalsax:
closing as this is out-dated and we have pivoted since the creation of the doc

PR #226, opened 2021-03-04
Add error handling for when conjurrc configuration is missing

sigalsax:
### What does this PR do?
Previously, when the conjurrc was lacking mandatory parameters, an unclear error was being returned. The error was just returning the name of the parameter missing, for example 'cert_file' without really explaining the problem or how to solve it. This becomes more apparently and visible because we changed the conjurrc parameter names from 
```
---
account: ...
appliance_url: ...
cert_file: ...
```
To:
```
---
cert_file: ...
conjur_account: ...
conjur_url: ...
```

#### Change log

#### Test coverage

#### Documentation


conjur/api/client.py
```diff
@@ -99,9 +97,10 @@ def __init__(self,
                               f"'cert_file': {loaded_config['ca_bundle']}}}")
             except CertificateVerificationException:
                 raise
-            # TODO add error handling for when conjurrc field doesn't exist
             except Exception as exc:
-                raise ConfigException(exc) from exc
+                raise ConfigurationMissingException("The conjurrc configuration file is either missing "
+                                                    "or missing parameters. Ensure that it exists and contains "
```
sigalsax, line 102: @InbalZilberman need you 👍  here

conjur/errors.py
```diff
@@ -50,3 +50,6 @@ class CertificateHostnameMismatchException(Exception):
     def __init__(self):
         self.message = MISMATCH_HOSTNAME_MESSAGE
         super().__init__(self.message)
+
+class ConfigurationMissingException(Exception):
+    """ Exception for when configuration is missing """
```
tzheleznyak, line 55: Should there be __init__ function inside the class?

conjur/config.py
```diff
@@ -29,25 +32,28 @@ class Config():
...
 
+        logging.debug(f"Fetching connection details from filesystem {config_file}")
         config = None
         with open(config_file, 'r') as config_fp:
             config = load(config_fp, Loader=Loader)
 
-        for config_field_name, attribute_name, mandatory in self.FIELDS:
-            if mandatory:
-                assert config_field_name in config
+        if config != None:
```
tzheleznyak, line 48:
```suggestion
        if not config:
```

tzheleznyak (dismissed):
Left 2 comments

conjur/api/client.py
InbalZilberman: @sigalsax How does the user make sure it exists? how about asking the user to run conjur init if needed?
has TW reviewed it?

sigalsax: reached out to Shuli.

The user makes it exist by reinitializing the client. I updated the message to
"The conjurrc configuration file is either missing, empty, or missing parameters. Reinitialize the client or ensure that the conjurrc contains 'conjur_account', 'conjur_url', and 'cert_file'."

conjur/errors.py
sigalsax: added.

conjur/config.py
```diff
@@ -29,25 +31,28 @@ class Config():
...
+        logging.debug(f"Fetching connection details from filesystem {config_file}")
         config = None
         with open(config_file, 'r') as config_fp:
             config = load(config_fp, Loader=Loader)
 
-        for config_field_name, attribute_name, mandatory in self.FIELDS:
-            if mandatory:
-                assert config_field_name in config
+        if not config:
+            for config_field_name, attribute_name, mandatory in self.FIELDS:
```
mbenita-Cyberark, line 48: could you please add comment here what and how are we checking? I didn't understood the FIELDS structure at first and what we are checking

conjur/config.py
```diff
@@ -29,25 +31,28 @@ class Config():
...
+                if mandatory:
+                    assert config_field_name in config
```
mbenita-Cyberark, line 50: I would throw specific error here and a different in the setattr.
this way you can inform if a value is mandatory or not

conjur/config.py
```diff
@@ -29,25 +31,28 @@ class Config():
...
 
+        logging.debug(f"Fetching connection details from filesystem {config_file}")
         config = None
         with open(config_file, 'r') as config_fp:
             config = load(config_fp, Loader=Loader)
 
-        for config_field_name, attribute_name, mandatory in self.FIELDS:
-            if mandatory:
-                assert config_field_name in config
+        if not config:
```
mbenita-Cyberark, line 47: haha. I would think about using `if config is None:` better explicit I think but this is negotiable .

tzheleznyak: @sigalsax Im very sorry @mbenita-Cyberark is right. I made it opposite

tzheleznyak: We can just do
`if config`

conjur/api/client.py
```diff
@@ -75,33 +72,37 @@ def __init__(self,
...
-                raise
-            # TODO add error handling for when conjurrc field doesn't exist
+                raise CertificateVerificationException(cause="The client was initialized without certificate verification, "
+                                        "even though the command was ran with certificate verification enabled.",
+                                        solution="To continue communicating with the server insecurely, run the command "
+                                        "again with ssl_verify = False. Otherwise, reinitialize the client.")
             except Exception as exc:
-                raise ConfigException(exc) from exc
+                raise ConfigurationMissingException("The conjurrc configuration file is either missing, empty, "
+                                                    "or missing parameters. Reinitialize the client or ensure that "
```
shulifink, line 103:
```suggestion
                                                    "or missing parameters. Reinitialize the client and make sure that "
```

shulifink (changes requested):

conjur/config.py
sigalsax: Yes this is very confusing logic. I will add a comment above. let me know if its more clear

conjur/config.py
sigalsax: as in two catches in client.py (this calling class) without the `else raise ConfigurationMissingException`?

except AssertionError
except ConfigurationMissingException

```
if mandatory:
  assert config_field_name in config, "THIS IS MANDETORY" # raises assertionerror

try:
   setattr(self, attribute_name, config[config_field_name])
   self._config[attribute_name] = getattr(self, attribute_name) 
except ConfigurationMissingException
```

conjur/api/client.py
sigalsax: updated

conjur/config.py
mbenita-Cyberark: Sorry I wasn't clear. i wouldn't raise assert but a dedicated exception like MissingParameterException.
Basically we should use assert to simplify verification specially in constructors [see here](https://wiki.python.org/moin/UsingAssertionsEffectively) but after that catch and throw a specific exception.

conjur/config.py
```diff
@@ -29,25 +31,29 @@ class Config():
...
         with open(config_file, 'r') as config_fp:
             config = load(config_fp, Loader=Loader)
 
-        for config_field_name, attribute_name, mandatory in self.FIELDS:
-            if mandatory:
-                assert config_field_name in config
-
-            setattr(self, attribute_name, config[config_field_name])
-            self._config[attribute_name] = getattr(self, attribute_name)
+        if config:
```
mbenita-Cyberark, line 47: we talked earlier , and wanted to reduce nested if's. 
what many code analayzer suggest is to reduce the use of else.
```suggestion
        if not config: raise ConfigurationMissingException
```
and the rest of the code could run outside of the "if" statement. I think it will be well suited here

conjur/config.py
```diff
@@ -29,25 +31,29 @@ class Config():
...
+            # TODO consider using load_from_file (ConjurrcData) instead of the following logic
+            # Sets the value of that FIELDS maps to with the values found in the conjurrc
+            for config_field_name, attribute_name, mandatory in self.FIELDS:
+                if mandatory and config_field_name not in config:
+                    raise InvalidConfigurationException
+                setattr(self, attribute_name, config[config_field_name])
```
mbenita-Cyberark, line 53: If we are going to refactor that file I think we should add a note here fo removing the setattr use.

conjur/config.py
```diff
@@ -29,25 +31,29 @@ class Config():
...
+                self._config[attribute_name] = getattr(self, attribute_name)
```
mbenita-Cyberark, line 54: same for getattr

mbenita-Cyberark (commented):
great work

conjur/config.py
sigalsax: i added the TODO note "# TODO consider using load_from_file (ConjurrcData) instead of the following logic" so I think it is clear enough that this whole section needs to be rethought. Do you still think a comment should made here even with this comment?

mbenita-Cyberark (approved):

PR #253, opened 2021-04-01
Fix bug for prompting user for url in insecure mode

sigalsax:
Previously when a user was running in insecure mode and not supplying the url, they would get an error "Invalid URL 'None/info': No schema supplied. Perhaps you meant http://None/info?". This fixes this by extracting the request for the URL from the certificate logic

conjur insecure init 
Returns error

conjur insecure init -u https://conjur-server
Return success

Resolves -

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/conjur/controller/init_controller.py
+++ b/conjur/controller/init_controller.py
@@ -54,9 +54,10 @@ class InitController:
             # or CA-signed, we will write the certificate on the machine
             self.write_certificate(fetched_certificate)
         else:
-            logging.warning('You chose to initialize the client in insecure mode. Reinitialize '
-                            'the client should you choose to communicate to the server securely')
-            self.conjurrc_data.cert_file=""
+            logging.warning('You chose to initialize the client in insecure mode. If you '
+                            'If you prefer to communicate with the server securely, you '
+                            'must reinitialize the client in secure mode.')
+            self.conjurrc_data.cert_file = ""
 
         self.get_account_info(self.conjurrc_data)
         self.write_conjurrc()
--- a/conjur/controller/init_controller.py
+++ b/conjur/controller/init_controller.py
@@ -54,9 +54,6 @@ class InitController:
             # or CA-signed, we will write the certificate on the machine
             self.write_certificate(fetched_certificate)
         else:
-            logging.warning('You chose to initialize the client in insecure mode. If you '
-                            'If you prefer to communicate with the server securely, you '
-                            'must reinitialize the client in secure mode.')
             self.conjurrc_data.cert_file = ""
 
         self.get_account_info(self.conjurrc_data)
--- a/conjur/controller/init_controller.py
+++ b/conjur/controller/init_controller.py
@@ -43,8 +43,10 @@ class InitController:
         Method that facilitates all method calls in this class
         """
         if self.conjurrc_data.conjur_url is None:
-            self.get_conjur_server_url()
-        formatted_conjur_url = self.format_conjur_server_url()
+            self.get_conjur_url()
+
+        formatted_conjur_url = self.format_conjur_url()
+        self.validate_conjur_url(formatted_conjur_url)
 
         if self.ssl_verify is True:
             fetched_certificate = self.get_server_certificate(formatted_conjur_url)
@@ -61,7 +63,7 @@ class InitController:
 
         sys.stdout.write("Successfully initialized the Conjur CLI\n")
 
-    def get_conjur_server_url(self):
+    def get_conjur_url(self):
         """
         Method to get the Conjur server URL if not provided
         """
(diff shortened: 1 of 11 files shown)


conjur/controller/init_controller.py
```diff
@@ -42,22 +42,28 @@ def load(self):
...
+
         if self.ssl_verify is True:
-            fetched_certificate = self.get_server_certificate()
+            fetched_certificate = self.get_server_certificate(formatted_conjur_url)
             # For a uniform experience, regardless if the certificate is self-signed
             # or CA-signed, we will write the certificate on the machine
             self.write_certificate(fetched_certificate)
         else:
+            logging.warning('You chose to initialize the client in insecure mode. Reinitialize '
+                            'the client should you choose to communicate to the server securely')
```
sigalsax, line 56: @InbalZilberman need your 👀  here. When we initialize the CLI we want to print a warning _once_ instead of an ugly
```
/Users/ssax/Documents/work/conjur-api-python3/venv/lib/python3.9/site-packages/urllib3/connectionpool.py:1013: InsecureRequestWarning: Unverified HTTPS request is being made to host 'sigalmastercli.aim-dev.conjur.net'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  warnings.warn(
```

on each command. Example:

Previously:
```
conjur --insecure init
Enter the URL of your Conjur server (use HTTPS prefix): https://sigalmastercli.aim-dev.conjur.net
/Users/ssax/Documents/work/conjur-api-python3/venv/lib/python3.9/site-packages/urllib3/connectionpool.py:1013: InsecureRequestWarning: Unverified HTTPS request is being made to host 'sigalmastercli.aim-dev.conjur.net'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  warnings.warn(
File /Users/ssax/.conjurrc exists. Overwrite? yes/no (Default: yes): y
```
now
```
conjur --insecure init
Enter the URL of your Conjur server (use HTTPS prefix): https://sigalmastercli.aim-dev.conjur.net
WARNING: You chose to initialize the client in insecure mode. Reinitialize the client should you choose to communicate to the server securely
...
File /Users/ssax/.conjurrc exists. Overwrite? yes/no (Default: yes): y
```

test/test_unit_client.py
```diff
@@ -128,15 +128,17 @@ def test_client_passes_default_account_to_api_initializer_if_none_is_provided(se
             url='http://myurl',
         )
 
+    @patch('conjur.wrapper.keystore_adapter.KeystoreAdapter.is_keyring_accessible', return_value=False)
     @patch('conjur.api.client.Api')
-    def test_client_performs_password_api_login_if_password_is_provided(self, mock_api_instance):
+    def test_client_performs_password_api_login_if_password_is_provided(self, mock_api_instance, mock_accessible):
```
eladkug, line 133: Can you elaborate the use of mock_accessible ?

sigalsax: mock_accessible means that the keyring isn't accessible and because it is a unit test it doesnt matter which flow we use (keyring or netrc). I use this so I can mock the load function that is called from the Client because for this test we are just testing that the API gets the parameters passed from the Client

conjur/controller/init_controller.py
sigalsax: @shulifink approval need here as well

conjur/controller/init_controller.py
```diff
@@ -42,22 +42,28 @@ def load(self):
         """
         Method that facilitates all method calls in this class
         """
+        if self.conjurrc_data.conjur_url is None:
+            self.get_conjur_server_url()
```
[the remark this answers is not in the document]
sigalsax: extracted the two (format/validate) to their own method because I need the URL formatted both for when the user inputs the URL and when they didn't

conjur/controller/init_controller.py
```diff
@@ -66,25 +74,40 @@ def get_server_certificate(self):
...
+        ensure it is provided as expected
+        """
+        if conjur_url.scheme != 'https':
             raise RuntimeError(f"Error: undefined behavior. Reason: The Conjur URL format provided "
                    f"'{self.conjurrc_data.conjur_url}' is not supported.")
 
+    # pylint: disable=line-too-long
+    def get_server_certificate(self, conjur_url):
+        """
+        Method to get the certificate from the Conjur endpoint detailed by the user
```
eranha, line 101: IMHO we can drop the `Method to` prefix

```
Gets the certificate from the specified conjur_url
Return tuple of certificate fingerprint and certificates chain or None in case conjurrc contains a certificate file
```

conjur/wrapper/http_wrapper.py
```diff
@@ -36,6 +37,7 @@ def invoke_endpoint(http_verb, endpoint, params, *args, check_errors=True,
     """
     This method flexibly invokes HTTP calls from 'requests' module
     """
+    urllib3.disable_warnings()
```
eranha, line 40: Is it OK to disable it permanently?

eladkug (commented):
LGTM from test perspective

conjur/wrapper/http_wrapper.py
sigalsax: we disable the ugliness and replace it with a more readable error that open shows up on init. See my description [here](https://github.com/cyberark/conjur-api-python3/pull/253/files#r605500983). WDYT?

conjur/controller/init_controller.py
```diff
@@ -66,25 +74,40 @@ def get_server_certificate(self):
...
-        url = urlparse(self.conjurrc_data.conjur_url)
+        return urlparse(self.conjurrc_data.conjur_url)
 
-        # TODO: Factor out the following URL validation to ConjurrcData class
-        # and add integration tests
-        if url.scheme != 'https':
+    def validate_conjur_url(self, conjur_url):
+        """
+        Method for validating the Conjur server URL to
+        ensure it is provided as expected
```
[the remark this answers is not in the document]
sigalsax: in essence renaming validate_conjur_url -> is_valid_scheme?
I don't think this will be the only validation in the future. If we think about all commands, IMO it is easier to follow a single pattern of "validation" instead of calling each validation function in each command differently. That way it is easier to read and is general enough so the reader doesnt get bogged down into the implementation details

conjur/controller/init_controller.py
```diff
@@ -42,22 +42,30 @@ def load(self):
         """
         Method that facilitates all method calls in this class
         """
+        if self.conjurrc_data.conjur_url is None:
+            self.get_conjur_url()
```
mbenita-Cyberark, line 46: suggest renaming for `prompt_for_conjur_url()` the name `get_conjur_url` implies of a function that return the url ant not storing it inside the class variables

conjur/controller/init_controller.py
sigalsax: @InbalZilberman @shulifink reminder

conjur/controller/init_controller.py
sigalsax: fixed!

conjur/controller/init_controller.py
shulifink: conjur --insecure init
Enter the URL of your Conjur server (use HTTPS prefix): https://sigalmastercli.aim-dev.conjur.net
WARNING: You chose to initialize the client in insecure mode. If you prefer to communicate with the server securely, you must reinitialize the client in secure mode.
File /Users/ssax/.conjurrc exists. Overwrite? yes/no (Default: yes): y

shulifink: BTW, when the user enters yes or no, do they have to type the whole word, or can they use y and n like you did?

sigalsax: yes there are shortcuts so y/Y/YES/yes and the same for "no"

sigalsax: updated

InbalZilberman (approved):
Thank you! approved

InbalZilberman (approved):
approved

PR #307, opened 2021-06-14
Add certificate information to .gitignore file

sigalsax:
### What does this PR do?
.key and .crt files are generated when we run our tests suite locally. These files are only relevant locally and only during the lifecycle of the test environment containers. That being said, they shouldn't be pushed. This commit adds these two file extensions (.crt/.key) to the gitignore file so that we don't push them accidentally.

Resolves -


royico (approved):
LGTM

PR #312, opened 2021-06-30
Update RHEL artifacts to tar.gz

orenbm:
### What does this PR do?
We now ship a 'tar.gz' archive for RHEL so we need to
update the guidelines

#### Change log

#### Test coverage

#### Documentation
- [x] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -248,7 +248,7 @@ The following section provides instructions on what is needed to perform a Conju
 
 Before each release the following tests will need to be performed:
 
-- On *each* platform we support (macOS, RHEL 7/8, Windows), copy over the compressed zip that holds the CLI executable, 
+- On *each* platform we support (macOS, RHEL 7/8, Windows), copy over the compressed archive that holds the CLI executable, 
   decompress it, and [run integration tests](#running-tests-outside-of-a-containerized)
   
 - Run the integration tests against the following different Conjur server environments from any platform you choose:
@@ -309,6 +309,8 @@ See the below section _How to create release artifacts_ for detailed information
 
 - Sign Windows executable
 
+- Sign RHEL 7/8 executable
+
 - Sign and notarize the ConjurCLI app for macOS 
 
 *Important!* The final artifacts that are delivered to the customer should be created from the main branch
@@ -323,9 +325,9 @@ For all OS types perform the following:
 #### RHEL 7/8
 
 1. Run `pyinstaller -F ./pkg_bin/conjur`. Once this is run, a `dist` folder will be created with the executable in it.
-1. Once an executable has been created, run `zip conjur-cli-rhel-7 conjur` (`zip conjur-cli-rhel-8 conjur` for RHEL 8 
-   machines) to zip the file.
-1. Sign the `zip` and add it as an asset in the release page.
+1. Once an executable has been created, run `tar cvf conjur-cli-rhel-7.tar.gz conjur` 
+   (`tar cvf conjur-cli-rhel-8.tar.gz conjur` for RHEL 8 machines) to archive the file.
+1. Sign the archive and add it as an asset in the release page.
 
 #### macOS
 
@@ -350,11 +352,11 @@ To copy files over from Windows VM to your local machine, use Remote Desktop red
   connection to see the changes.
 1. Drag the executable/zip to the shared folder. You should now see it on your local machine.
 
-The zips should be called the following:
+The archives should be called the following:
 
 ```
-conjur-cli-rhel-7.zip
-conjur-cli-rhel-8.zip
+conjur-cli-rhel-7.tar.gz
+conjur-cli-rhel-8.tar.gz
 conjur-cli-windows.zip
 conjurcli.dmg
 ```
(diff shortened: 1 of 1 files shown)


Tovli (approved):
LGTM
Thanks

CONTRIBUTING.md
```diff
@@ -323,9 +325,9 @@ For all OS types perform the following:
 #### RHEL 7/8
 
 1. Run `pyinstaller -F ./pkg_bin/conjur`. Once this is run, a `dist` folder will be created with the executable in it.
-1. Once an executable has been created, run `zip conjur-cli-rhel-7 conjur` (`zip conjur-cli-rhel-8 conjur` for RHEL 8 
-   machines) to zip the file.
-1. Sign the `zip` and add it as an asset in the release page.
+1. Once an executable has been created, run `tar cvf conjur-cli-rhel-7.tar.gz conjur` 
+   (`tar cvf conjur-cli-rhel-8.tar.gz conjur` for RHEL 8 machines) to archive the file.
+1. Sign the archive and add it as an asset in the release page.
```
sigalsax, line 330: "add them both as assets" because we are taking about two tars

CONTRIBUTING.md
```diff
@@ -350,11 +352,11 @@ To copy files over from Windows VM to your local machine, use Remote Desktop red
   connection to see the changes.
 1. Drag the executable/zip to the shared folder. You should now see it on your local machine.
 
-The zips should be called the following:
+The archives should be called the following:
```
sigalsax, line 355: what you say archives are you suggesting only the RHEL 7 / 8 archives or all of the deliverables

I would change word to be "deliverables"

CONTRIBUTING.md
sigalsax: let's add something about how we need to push the .sigs as well for signature validation

orenbm: when we run `pyinstaller -F ./pkg_bin/conjur` does it create the same executable for RHEL 7 and RHEL 8? if so, why do even have 2 artifacts? if not - how do you create one per platform?

orenbm: updated the doc according to what i understand. please review

CONTRIBUTING.md
orenbm: changed

CONTRIBUTING.md
sigalsax: you need to spin up two different machines for RHEL 7 and RHEL 8 and then run the `pyinstaller -F ./pkg_bin/conjur` on each. Pyinstaller is a tool that, amongst other things, allows us to pack our CLI into a single executable. 

The limitation of Pyinstaller that you need to compile down the exec on the OS that it is intended to run on. For example, an executable compiled on macOS won't run on Windows. For RHEL, it is interesting because I have seen that an executable compiled on RHEL 8 can work on RHEL 7 but **not** vice versa. "So why not just deliver a RHEL 8 exec?" because although it _seems_ to work, this might not always hold in later versions and just to be on the same side and not OS-dependent, we deliver two.

CONTRIBUTING.md
```diff
@@ -323,9 +325,13 @@ For all OS types perform the following:
 #### RHEL 7/8
 
 1. Run `pyinstaller -F ./pkg_bin/conjur`. Once this is run, a `dist` folder will be created with the executable in it.
```
sigalsax, line 327: I know this wasn't your doing but I feel like it's not clear that its two different machines. Can you update this first sentence to:
```
Run `pyinstaller -F ./pkg_bin/conjur` on the different RHEL 7 and RHEL 8 machines.
```

CONTRIBUTING.md
```diff
@@ -323,9 +325,13 @@ For all OS types perform the following:
...
-1. Once an executable has been created, run `zip conjur-cli-rhel-7 conjur` (`zip conjur-cli-rhel-8 conjur` for RHEL 8 
-   machines) to zip the file.
-1. Sign the `zip` and add it as an asset in the release page.
+1. Once an executable has been created, archive the file for RHEL 7 and RHEL 8 platforms, using the following commands:
+  1. `tar cvf conjur-cli-rhel-7.tar.gz conjur`
```
sigalsax, line 329: why do we need `c` instead of `x`? We don't want to create a new folder, instead we just want to extract the contents of the tar to the directory where that command was run

sigalsax: on my RHEL 7 machine, I see the following error
```
$ tar cvf conjur-cli-rhel-7.tar.gz conjur
tar: conjur: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```

orenbm: i don't follow. you have the tar and you want to extract it? if so - you need to run `tar xvf` and not `tar xvf`

sigalsax:
```
tar xvf conjur-cli-rhel-7.tar.gz

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

CONTRIBUTING.md
orenbm: TIL

CONTRIBUTING.md
sigalsax: seems the exec was corrupted. fixed with @orenbm

sigalsax (approved):
LGTM! thanks :)

sigalsax (approved):

PR #316, opened 2021-07-04
Use admin password from ConjurOps

### LINKED ISSUE #301: Flaky appliance deployment in RHEL on password complexity error
## Summary
The CLI/SDK pipeline is failing on RHEL8  Integration Tests because of a password complexity issue when deploying the appliance. The randomly generated password does not mean the complexity requirement and this should be changed to pull the secret from the ConjurOps server using summon.
Screen Shot 2021-05-24 at 17 26 18

## Steps to Reproduce
1. Go to project's jenkins pipeline
2. Run (re-run) the pipeline
3. You should se the error pictured above. If you don't re-run the pipeline until you do

## Expected Results
Pull secret from ConjurOps using summon so we won't see this flakiness

## Actual Results (including error logs, if applicable)
```
error: Admin password The password you have chosen does not meet the complexity requirements. Choose a password that includes: 12-128 characters, 2 uppercase letters, 2 lowercase letters, 1 digit, and 1 special character.
```

## Reproducible
   * [ ] Always 
   * [x] Sometimes
   * [ ] Non-Reproducible
   
## Version/Tag number
-

## Environment setup
RHEL 8

## Additional Information
-

orenbm:
### What does this PR do?
We prefer to use the admin password from ConjurOps
instead of generating one. This helps us also in the
complexity and also in referring to the password as
a secret

Resolves #301

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/bin/test_integration_rhel8
+++ b/bin/test_integration_rhel8
@@ -25,19 +25,7 @@ docker-compose up openssl
 echo "Starting Conjur..."
 docker-compose up -d conjur-appliance
 
-function _random_password() {
-  { </dev/urandom LC_ALL=C grep -ao '[A-Za-z0-9]' \
-        | head -n$((RANDOM % 12 + 9))
-    echo "@1"   # add a special char and digit.
-  } \
-    | shuf \
-    | tr -d '\n'
-}
-
 echo "Configuring Conjur..."
-set +x
-export ADMIN_PASSWORD=$(_random_password)
-set -x
 export DEBUG=$DEBUG
 export CONJUR_AUTHN_API_KEY=$ADMIN_PASSWORD
 export SERVER_MODE="appliance"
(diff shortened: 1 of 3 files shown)


bin/test_integration_rhel8
```diff
@@ -35,7 +35,6 @@ function _random_password() {
 }
 
 echo "Configuring Conjur..."
-export ADMIN_PASSWORD=$(_random_password)
```
sigalsax, line 38: you can also get rid of `_random_password` function since this is the only place we use it

docker-compose.yml
```diff
@@ -104,7 +104,6 @@ services:
       TEST_ENV: "True"
       SERVER_MODE: Null
       LOGIN: Null
-      ADMIN_PASSWORD: Null
```
sigalsax, line 107: shouldnt this remain and just be `ADMIN_PASSWORD: $ADMIN_PASSWORD`

?

sigalsax (approved):

PR #334, opened 2021-08-22
Fix host example from command it does not support

sigalsax:
### What does this PR do?
The host command had a bug under 'examples' section where it was displaying 'change-password' instead of 'rotate-api-key' when host does not support the change-password command. This commit fixes that and also fixes the broken main build on pylint encoding suggestion

Bug:
Screen Shot 2021-08-22 at 8 42 56
Fix:
Screen Shot 2021-08-22 at 8 43 15

Resolves -

#### Change log

#### Test coverage

#### Documentation


royico (approved):
LGTM

PR #335, opened 2021-09-09
Add SDK refactor design document

sigalsax:
### What does this PR do?
This commit adds the SDK refactor design doc


sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,351 @@
...
+
+| Resource                  | Link                                      |
+| ------------------------- | ----------------------------------------- |
+| Zen of Python (the Bible) | https://www.python.org/dev/peps/pep-0020/ |
+
+
+
+### Motivation
+
+While GAing the CLI, it was realized that the SDK has the following limitations, making the code not easily readiable and hard to make additions:
```
orenbm, line 30:
```suggestion
While GAing the CLI, it was realized that the SDK has the following limitations, making the code not easily readable and hard to make additions:
```

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,351 @@
...
+
+    `client = conjur.Client(url="https://someserver", account="cucumber", password="123", ...)`.
+
+2. The *Client* constructor is called, ensuring all configuration and credential information needed to make a request have been passed in. If not, it fetches the missing information in a discovery-like manner. For example, for `login`, if a Keystore is not available, it will attempt to fetch credentials from the netrc. At this point the *API* is also initalized.
+
+3. *Client* is now initalized and the user can now make a request.
+
+   `list_values = client.list()`
+
+4. Once a request is made, the *Client* makes a call to *API* which handles the building of the request, merging paramters together, understanding which endpoint to use, etc.
```
orenbm, line 60:
```suggestion
4. Once a request is made, the *Client* makes a call to *API* which handles the building of the request, merging parameters together, understanding which endpoint to use, etc.
```

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,351 @@
...
+
+6. A response is returned to the *HTTPWrapper*, handing it back to the calling *API* which formats the response to the desired output.
+
+7. *API* returns the formatted response to the *Client* and the *Client* returns it to the user. 
+
+
+
+### Limitations in current SDK Flow
+
+This section will be divided according to the three areas in the code that need to most attention (Client, Api, and Commands). In each section, I will outline their limitations, the consequences of their limitations, and provide a proposal for how to address them.
```
orenbm, line 72:
```suggestion
This section will be divided according to the three areas in the code that need the most attention (Client, Api, and Commands). In each section, I will outline their limitations, the consequences of their limitations, and provide a proposal for how to address them.
```

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,351 @@
...
+          ...
+        else
+          ...     
+```
+
+See Client [code](https://github.com/cyberark/conjur-api-python3/blob/main/conjur/api/client.py#L51) for reference.
+
+1. Multiple responsibilities and not easily expandable.
+
+   a. The constructor consumes 92 lines and handles fetching configuration and credential data, validating the input, configuring API calls, error handing, etc. This makes impacts code readability and should be broken up.
```
orenbm, line 109:
```suggestion
   a. The constructor consumes 92 lines and handles fetching configuration and credential data, validating the input, configuring API calls, error handling, etc. This makes impacts code readability and should be broken up.
```

orenbm (commented):
Thanks a lot @sigalsax ! this design looks great!

Can you address please backwards compatibility? are we breaking it? 
I would also get some Python expert (@tzheleznyak ?) to take a look as i'm not very familiar.

Can you also write a delivery plan with breaking and estimations?

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,390 @@
...
+| 3    | Generic Config/Creds object                     | Configuration()<br />Credentials()<br />and we will discover where the connection details live based on environment's context | - Easy maintainability. No change in user's code as we build new ways to fetch config and creds<br />- Aligns with CLI UX of Config/Cred discovery | - Implicit and the flow we choose is not clear               |
+|      | A single Config/Cred object with defining param | Configuration(action=file) - .conjurrc<br />Credential(action=keystore) | - Explicit<br />- Minimal code changes should the user choose a different flow<br />- Strong drive in Python to be as explicit as possible | - Requires changes in code if user wants to adopt another retrieval method<br />- Conflicting experience with CLI. The CLI preforms a discovery on Config/Creds retrieval type |
+
+**Decision:** Option #4.
```
tzheleznyak, line 155: Why option 4 and not 2 . I agree with not choosing 1 and 3?

tzheleznyak:
there is a factory for cred objects: why the objects called Configuration for file and Credentials for keystore. Cant they have more similar names like KeystoreCreds and FileCreds for example ?

sigalsax:
@tzheleznyak to start, the design is very much an opinion and open to go in any direction we believe in and want to go down. I made decisions based on what I thought was most intuitive when thinking about UX and implementation. That being said, I am not bound to any idea so happy you are asking these questions

Now to your questions,

The idea is to have `Configuration(action=file)` so any change requires the least amount of effort by the user. Instead of changing the object, they change a simple param. In essence the two flows are the same thing because in the end, in the background, we are creating a KeystoreCreds and a FileCreds. There is just an extra layer of abstraction called Configuration/Credential factory so that if we have common logic, if can happen in that factory

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,390 @@
...
+
+We can handle the separation in one of the following ways:
+
+|      | Solution                                        | Description                                                  | Pros                                                         | Cons                                                         |
+| ---- | ----------------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| 1    | Keep as is                                      |                                                              | - Doesn't break current user apps<br />- Seems to be the direction the industry is going | - 92 lines and 9 params so far makes it unweidy<br />- Constructor is overloaded, handling multiple responsibilities<br />- Doesn't take advantage of OOP |
```
abrahamko, line 150: > Seems to be the direction the industry is going

Please give some proof to this statement, otherwise it is hard to evaluate it.

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,390 @@
...
+
+   b. We can expect changes in the following areas in future iterations:
+
+   ​	i. Method of fetching configuration. Currently, we support fetching configuration details from the conjurrc file but this can change in the future and we will support methods in the future (via environment variables, etc).
+
+   ​	ii. Method of fetching credentials (via the netrc file or keystore). Currently, we support fetching credential data from the keystore or the netrc file.
+
+   Because of all these possible changes, we need to encapsulate them so that when changes arise, they are contained, without effecting the system.
+
+For these reasons, I propose we extract the configuration and credential information to their own objects to be created by the user before the Client is called. In those classes, all the collection and validations will be performed. As a result, the Client constructor will not be forced to be in charge of handle multiple responsibilities, the user will have to explicitly define the and the areas of volatility will be contained.
```
abrahamko, line 144: What *is* the `Client` class responsibility?
It is not obvious, at least to me...

mbenita-Cyberark (commented):
It could go to many directions let's talk?

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+**Decision:** Option #4. 
+
+The flow for configuration and credential will be determined *explicitly* by enabling the user to pass in the flow of their choosing via a parameter to the object during instantiation. 
+
+This functionality will be implemented using the [Factory pattern](https://refactoring.guru/design-patterns/factory-method). Each factory-generated class will have its own implementation and validation mechanism. This will make the implementation easy to maintain as this makes it easy to swap algorithms used inside an object during runtime. That way, we can isolate the implementation details from the *Client*. If we choose to implement a new strategy, for example fetching values from environment values, this will not require a change in the Client.
```
abrahamko, line 159: Is it true that this means no discovery?
We can consider adding a static method to discover: `discoverCredentialStore()`.
It will look for the relevant credential store, in the same way it is done today.

sdk_refactor_design_proposal.md
mbenita-Cyberark: After giving at another look I agree 2 is better. we can expose credentialProviders as part of our module so the user can create CredentialData object from file or keystore. but the Client should get as a parameter CredentialsData object

sdk_refactor_design_proposal.md
mbenita-Cyberark, line 159: Regarding the Client structure I suggested something like that:
` class Client:

    def __init__(self, credentials, connection_data, **kwargs):
        self.credentials = credentials
        self.connection_data = connection_data
        self._variable = None
        self._policy = None`
        `    @property
    def variable(self):
        if self._variable is None:
            self._variable = Variable(self.credentials, self.connection_data)
        return self._variable`
        `class Variable:
    def __init__(self, credentials, connection_data):
        self.credentials = credentials
        self.connection_data = connection_data

    def get_one(self, variable_id):
        pass

    def get_many(self, ids):
        asyncio.run(self.get_many_async())

    async def get_many_async(self):
        pass`
        
The way we going to use this interface is by running 
        `    cred = CredentialsData()
    connectionData = ConjurrcData()
    client = Client(cred, connectionData)
    client.variable.get_many([1, 2, 3])`

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+We will *not* create the conjurrc and the netrc for the user as is true in the current state of the SDK. 
+
+We will *not* support partial states as is true in the current SDK. For example, we will not support the flow if a user provides `client.Configuration(action=file, url="https://someserver", account="", cert_file="/some/certfile")`. In this flow we will fail with relevant error.
+
+
+
+##### API and commands
+
+The following code excerpt was taken from the current implementation of *Client* and *API* classes:
```
mbenita-Cyberark, line 220: regarding the API I suggest going with different approach. implement webapi service as done in other clientside applications.
something like 
`class RestAipProvider:

    def __init__(self, url, api_key=None, ssl_verify=None):
        self.url = url
        self.api_key = api_key
        self.ssl_verify = ssl_verify

    async def authenticate(self, account, login, pasrams=None):

        endpoint = f"{self.url}/authn/{account}/{login}/authenticate"
        async with aiohttp.ClientSession() as session:
            async with session.post(endpoint, data=pasrams) as resp:
                return await resp.json()


    async def get_variable(self, account, kind, identifier, params=None):
        endpoint = f"{self.url}/secrets/{account}/{kind}/{identifier}"
        async with aiohttp.ClientSession() as session:
            async with session.get(endpoint, data=params) as resp:
                await resp.json()
        return resp`
Where each function exposes serverside endpoint and return a promise with the server data response.
This layer should not try to transform the data
The benefits for this approach
1. webapi is scoped to one layer
2. very easy to extand\ add new apis
3. short and very easy to test functions
4. 100% async (we don't have to but I think we do specially for the sdk)

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+![CurrentSDKFlow](design/assets/CurrentSDKFlow.png)
+
+At a high level,
+
+1. The user initializes the *Client* with details needed to connect to the *Conjur server*
+
+    `client = conjur.Client(url="https://someserver", account="cucumber", password="123", ...)`.
+
+2. The *Client* constructor is called, ensuring all configuration and credential information needed to make a request have been passed in. If not, it fetches the missing information in a discovery-like manner. For example, for `login`, if a Keystore is not available, it will attempt to fetch credentials from the netrc. At this point the *API* is also initalized.
```
uCatu, line 67: What is the flow if partial arguments are passed by method? Is it still going into discovery mode?
Do we think about expanding it to environment variables like AWS CLI has?

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+3. *Client* is now initalized and the user can now make a request.
```
uCatu, line 69: Are we planning to have a TTL for this initialization?

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+   `list_values = client.list()`
+
+4. Once a request is made, the *Client* makes a call to *API* which handles the building of the request, merging parameters together, understanding which endpoint to use, etc.
+
+5. The *API* calls the *HTTPWrapper* which escapes the parameters and invokes the endpoint to the *Conjur Server*.
```
uCatu, line 75: What methods are we using to sanitize those parameters? Can you give an example for before and after?

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+                 url: str = None):
+      ...
+      if not url or not login_id or (not password and not api_key):
+        # load in Configuration details (if not provided)
+        # load in Credential details (if not provided)
+        # instantiate API objects according to the different params provided
+        ...
+        if api_key:
+          ...
+        elif password:
```
uCatu, line 112: We are considering just one of those two arguments but not saying which we accepted, for example:
`client(api_key:"123", password: "123")`
will work, but no indication for client what is redundant, I don't say that we need to give back indication, but to be sure if we want to create a distinction between api_key and password arguments. 

Also, what will be the behavior when debug if False but http_debug is True ? 
I would also like to think about the order of those argument input as debug/http_debug should be at the end of the list

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+# needed to connect to the Conjur server
+class ConfigurationModel(action = None,
+                         url = None,
+                         account = None,
+                         ca_bundle = None,
+                         ssl_verify = True):
+
+# configuration factory deciding which flow to run based on user input
+class Configuration(action=None, 
+                    ssl_verify=None)
```
uCatu, line 180: ssl_verify is not in use?

sdk_refactor_design_proposal.md
eranha: The downside that it violates SRP

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,390 @@
...
+| ---- | ----------------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| 1    | Keep as is                                      |                                                              | - Doesn't break current user apps<br />- Seems to be the direction the industry is going | - 92 lines and 9 params so far makes it unweidy<br />- Constructor is overloaded, handling multiple responsibilities<br />- Doesn't take advantage of OOP |
+| 2    | Config/Creds objects for *each* supported flow  | ConfigurationFromFile()  - .conjurrc<br />ConfigurationFromEnv() - ENV vars<br />CredentialsFromFile() - netrc<br />CredentialsFromKeystore() - keystore | - Explicit, the flow the user chooses is very clear<br />- Security is up to user's the discretion | - Conflicting experience with CLI. The CLI performs a discovery on Config/Creds retrieval type<br />- Requires changes when a user wants to choose another flow |
```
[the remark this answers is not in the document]
Tovli: I would follow this approach but would create a general configuration interface that responsible for handling configurations, which credentials is part of. so you basically have an object to represent a configuration:
```
{
  ConfigName: string,
  ConfigValue: string,
  IsSecret: boolean
}
```
Then you just do `config.getconfig(name)` and don't care where it get it from. 
The configuration interface can have a documented flow (first check file, then keystore if secret or env var if not)

sdk_refactor_design_proposal.md
Tovli: @mbenita-Cyberark - can you explain why 2 is better? we can expose a provider also in no. 4.
IMHO we don't need this extensibility point as I can't see a flow where you want the consumer of the config control the internal of where to get it from. It is a leakage of the DAL into a higher layer IMO

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+| 3    | Generic Config/Creds object                     | Configuration()<br />Credentials()<br />and we will discover where the connection details live based on environment's context | - Easy maintainability. No change in user's code as we build new ways to fetch config and creds<br />- Aligns with CLI UX of Config/Cred discovery | - Implicit and the flow we choose is not clear               |
+|      | A single Config/Cred object with defining param | Configuration(action=file) - .conjurrc<br />Credential(action=keystore) | - Explicit<br />- Minimal code changes should the user choose a different flow<br />- Strong drive in Python to be as explicit as possible | - Requires changes in code if user wants to adopt another retrieval method<br />- Conflicting experience with CLI. The CLI preforms a discovery on Config/Creds retrieval type |
+
+**Decision:** Option #4. 
+
+The flow for configuration and credential will be determined *explicitly* by enabling the user to pass in the flow of their choosing via a parameter to the object during instantiation.
```
Tovli, line 157: Why would we want to do that? The BL want config (which creds are part of), why does it matter if those are taken from a file or the keystore?

sdk_refactor_design_proposal.md
Tovli, line 159: It actually seems that you are describing the [strategy pattern](https://refactoring.guru/design-patterns/strategy) which is very similar to factory (note that factory is to encapsulate creation while strategy encapsulate behavior).
But I still argue that we don't need to pass the retrieval method so no need for either here

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+# data object representing the configuration data 
+# needed to connect to the Conjur server
+class ConfigurationModel(action = None,
+                         url = None,
+                         account = None,
+                         ca_bundle = None,
+                         ssl_verify = True):
+
+# configuration factory deciding which flow to run based on user input
+class Configuration(action=None,
```
Tovli, line 179: If this implement factory then why not call it `configurationFactory` ?

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+We will *not* support partial states as is true in the current SDK. For example, we will not support the flow if a user provides `client.Configuration(action=file, url="https://someserver", account="", cert_file="/some/certfile")`. In this flow we will fail with relevant error.
+
+
+
+##### API and commands
+
+The following code excerpt was taken from the current implementation of *Client* and *API* classes:
+
+```python
+class Client()
```
Tovli, line 223: The name `client` is very confusing to me... client of what?

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+  def __init__(...):
+    self._default_params={...}
+    self._api = Api(...)
+
+  def list(self, list_constraints: dict = None) -> dict:
+      return self._api.resources_list(list_constraints)
+
+    
+class Api()
```
Tovli, line 232: Why is this called `API` ? each class communicates with an API.... What is the responsibility ? seems to me that you have here a `ConjurServerResourceAccess` or maybe `ConjurAccess`? seems like a DAL to fetch data from conjur... 
A good practice is to think of Conjur as the CLI/SDK database... what would you call the object that comm. with the DB? `API` ? it surely communicate over an API :)

sdk_refactor_design_proposal.md
Tovli: Note that in this method you are actually creating a [facade](https://refactoring.guru/design-patterns/facade). Not saying this make this idea not good but it also means you are leaking the other system terminology (conjur server) into your application. The up side IMO is clear understanding of the calls we make

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+The *API* class handles multiple responsibilities. The API class handles the building of the request based on the types of information provided by the user *and* the invoking of the call *and* formatting the response.
+
+2. Areas of volatility are not encapsulated.
+
+We can add commands, change the way we present a certain command's output, and/or the types of parameters can manipulate the overall flow. Each command has its own associated function in `client.py` and `api.py` so each time we want to add a new command, we need to re-open multiple classes and add the desired functionality. This overcrowds the *Api*, allowing command-specific logic to spill into a very general class. 
+
+##### Proposed flow
+
+For these reasons, I propose we extract the command-specific logic into its own class. That way, the the *Api* class will be exposed to the specific details of a command, how to package the request, how to make the call, what format the request should be in. This works to isolate the areas of volatility should we require to support different use cases for a command.
```
Tovli, line 287: Seems a bit over-engineering to me. you usually encapsulate logic that change together. To me it seems that all three stages (build, execute, format) will change together. I would still search for a library that can help doing the generic work of sending the request and formatting the response

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+
+
+### Full proposed SDK flow
+
+![SDKReactorFlow](design/assets/SDKReactorFlow.png)
+
+At a high level,
+
+1. The user will initalize a Configuration and Credential object, *separate* from the Client
```
Tovli, line 344: See my comment reg. having a documented fetching logic so we won't need to initialize the configuration object

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+
+7. The *ListAPI* will then make a request to the *APIClient* which will handle the escaping of parameters, build headers, decide to run in a secure or unsecure flow, etc.
+
+8. *APIClient* makes a request to the *Conjur server* and returns the response to be formatted.
+
+9. The formatted response is returned to the user.
+
+
+
+### Backwards compatibility
```
Tovli, line 370: I would add a sentence about the versioning. what will happen after these changes?

Tovli (commented):
Very nice work
comments added

sdk_refactor_design_proposal.md
sigalsax: @mbenita-Cyberark I intended the user to be able to create a CredentialObject/ConfigurationObject regardless. See under "Proposed flow", the following sentence:

> For both Configuration and Credential flows, we will still support the ability for users to build these objects themselves

sdk_refactor_design_proposal.md
sigalsax: To name a few bigger projects
- https://github.com/boto/boto3/blob/develop/boto3/session.py#L48
- https://github.com/Miserlou/Zappa/blob/master/zappa/core.py#L243

sdk_refactor_design_proposal.md
sigalsax: Before or after my proposal?

Before, it happened 1. building config and cred objects 2. their validations 3. navigating requests according to the params it provided 5. and smart enough to know what to do when a param was missing

After my proposal, I plan for it to be an orchestrator, blind to the credential/configuration and command implementation details. It just gets inputs and shells out general instructions that will change according to the types of flow the user runs

sdk_refactor_design_proposal.md
sigalsax: See the bottom under "Proposed flow" where I mention:

> We will *not* create the conjurrc and the netrc for the user as is true in the current state of the SDK. 

> We will *not* support partial states as is true in the current SDK. For example, we will not support the flow if a user provides `client.Configuration(action=file, url="https://someserver", account="", cert_file="/some/certfile")`. In this flow we will fail with relevant error.

I don't think we should go into discovery mode because it is an endless black hole. If a user provided username but not password, we would go into discovery mode and look for the password in the keystore and netrc. Let's say keystore has it and we find it there. Then realize they didn't provide url so again we perform this discovery again but this time it is _not_ in the keystore, rather in the netrc. So you see we have this weird state where we are pulling variables from different available locations

> Do we think about expanding it to environment variables like AWS CLI has?

Definitely on our radar! We won't be adding this functionality though as part of the refactor. In the refactor our goal is to think of the project and lay it out in a way that adding this functionality won't be hard

sdk_refactor_design_proposal.md
sigalsax: This is the old flow (what you are referencing) which doesn't have an expiration but I think this would be really great if we introduced this.
Currently in the client, we authenticate the user to get the access token for them when they provide the Credentials and Config data). We can definitely separate this logic so that they have to run something like `client.init()`, `client.login()`, `client.authenticate()`
WDYT?

sdk_refactor_design_proposal.md
sigalsax: Not sure yet. Currently though we escape. See [here](https://github.com/cyberark/conjur-api-python3/blob/main/conjur/wrapper/http_wrapper.py#L46)

Any suggestions on what to introduce?

sdk_refactor_design_proposal.md
sigalsax: IMO we should make the distinction between password and API key because with password, we need to perform a login to get the API key. 

> We are considering just one of those two arguments but not saying which we accepted

We can definitely document precedence in our docs

> Also, what will be the behavior when debug if False but http_debug is True ?
debug = true, see verbose logging of the process flow. debug = false, you won't see any logs at all printed to STDOUT. Only error or success. 

http_debug = true, we can thoroughly debug HTTP requests and HTTP responses

sdk_refactor_design_proposal.md
sigalsax: should be defaulted to True, updating..

sdk_refactor_design_proposal.md
```diff
@@ -0,0 +1,392 @@
...
+The flow for configuration and credential will be determined *explicitly* by enabling the user to pass in the flow of their choosing via a parameter to the object during instantiation. 
+
+This functionality will be implemented using the [Factory pattern](https://refactoring.guru/design-patterns/factory-method). Each factory-generated class will have its own implementation and validation mechanism. This will make the implementation easy to maintain as this makes it easy to swap algorithms used inside an object during runtime. That way, we can isolate the implementation details from the *Client*. If we choose to implement a new strategy, for example fetching values from environment values, this will not require a change in the Client.
+
+##### Proposed flow
+
+```python
+config = client.Configuration(action=file)
+creds = client.Credental(action=keystore)
+client. conjur.ConjurClient(config, creds, debug, ssl_verify)
```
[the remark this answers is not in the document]
sigalsax: a context object as in? So it will look like
```
Context {
ConfigurationObject
CredentialsObject
debug
ssl_verify
}
```
?

sdk_refactor_design_proposal.md
sigalsax: I agree. You are referencing the current SDK logic. 

I wanted to update this to `ConjurClient`. WDYT?

sdk_refactor_design_proposal.md
sigalsax: This is referencing the current SDK. The one I hope to change :)
In my proposal design, I wanted to create an AccessLayer for each of our resources so it would be ListApi, VariableApi, etc. See under _API and commands_ 

I can change this to ListAccess and VariableAccess if that makes it clearer. Let me know which one you suggest would be clearer

sdk_refactor_design_proposal.md
sigalsax: Can you provide an example of what you mean by encapsulating the logic that changes together?
I thought that this was what I was doing by encapsulating the 3 in their own resource class

sdk_refactor_design_proposal.md
sigalsax: For the sake of being as explicit as possible so the user doesn't have to guess which context their environment is running in. 
Do you suggest, performing a discovery is better? For example have the user provide a general `creds = conjur.Credentials(username, password, url), and we search for them through mediums we support (netrc, keystore)?

sdk_refactor_design_proposal.md
mbenita-Cyberark: We don't have to use the same terminology..
The function in this API knows the server endpoint but exposes it own functionality. I agree that in most of the cases it will be straightForward but it will be our choice to do so. BTW I don't think it's validate the SRP as the server and the CLI are two different programs. 
Another thing: working with tools like protobuff or bond would have auto generate this layer for you, making me believe this is a step to the right direction

sdk_refactor_design_proposal.md
mbenita-Cyberark: @Tovli the keyStoreCredentialProvider can do CRUD operations on the keystore this is powerful, cross-platform and safer tool to use since it lowers the chance the user will write the credentials hard-coded in his/her app. But maybe we don't to expose this since it's another interface we need to maintain.
So I'm in favor to simply expose the user a `Credentials` and `Configurations` classes and let the user construct them in his own way.
Please note that python doesn't support constructor overloading. so supporting both `Credential(action=keystore) ` and `Credential(....) ` api will be pretty messy since it requires *`*kwargs` that we wanted to avoid.

sdk_refactor_design_proposal.md
mbenita-Cyberark: I don't think the sdk should support .netrc or .conjurrc files. it is up for the user to provide this data to the SDK at runtime.
.netrc is not safe and I don't think we should keep supporting it in the sdk.

PR #353, opened 2021-10-28
Log full error on policy load failure

hzur:
### Desired Outcome

When failing in load policy we'll print the exact error send from the server.

### Implemented Changes

On raising error we'll check the http error response text in addition to the error itself.

### Connected Issue/Story

Resolves #[relevant GitHub issue(s), e.g. 76]

[ONYX-12413](https://ca-il-jira.il.cyber-ark.com:8443/secure/RapidBoard.jspa?rapidView=1475&view=detail&selectedIssue=ONYX-12413)

### Definition of Done
*At least 1 todo must be completed in the sections below for the PR to be
merged.*

#### Changelog

- [ ] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

- [ ] This PR includes new unit and integration tests to go with the code
  changes, or

#### Documentation

#### Behavior

#### Security

- [ ] There are no security aspects to these changes

### DIFF
--- a/conjur/controller/policy_controller.py
+++ b/conjur/controller/policy_controller.py
@@ -35,5 +35,8 @@ class PolicyController:
         except requests.exceptions.HTTPError as http_error:
             if http_error.response.status_code == 422:
                 raise InvalidFormatException(f"{http_error}. The policy was empty or its contents "
-                                             "were not in valid YAML.") from http_error
-            raise
+                                             f"were not in valid YAML."
+                                             f"Error: {http_error.response.text}") from http_error
+            raise requests.exceptions.HTTPError(f"{http_error}."
+                                         f"Error: {http_error.response.text}",
+                                         response=http_error.response) from http_error
--- a/test/test_integration_policy.py
+++ b/test/test_integration_policy.py
@@ -117,6 +117,31 @@ class CliIntegrationPolicy(IntegrationTestCaseBase):  # pragma: no cover
                   str(mock_log.output))
 
     @integration_test()
+    def test_policy_bad_syntax_error_user_recieve_detailed_error(self):
+        policy_with_bad_syntax = "- ! user bad syntax"
+        with redirect_stderr(self.capture_stream):
+            output = utils.load_policy_from_string(self, policy_with_bad_syntax, exit_code=1)
+        self.assertIn("{\"error\":{\"code\":\"validation_failed\",\"message\":", output, 
+                        "Could not find detailed error in the output."
+                        "Expected to find:"
+                        "{\"error\":{\"code\":\"validation_failed\",\"message\":"
+                        "Result:"
+                        f"{output}")
+
+    @integration_test()
+    def test_policy_load_bad_branch_recieve_detailed_error(self):
+        with redirect_stderr(self.capture_stream):
+            output = self.invoke_cli(self.cli_auth_params,
+                           ['policy', 'load', '-b', 'wrong-branch', '-f', 
+                           self.environment.path_provider.get_policy_path("resource")], exit_code=1)
+        self.assertIn("{\"error\":{\"code\":\"not_found\",\"message\":", output, 
+                        "Could not find detailed error in the output."
+                        "Expected to find:"
+                        "{\"error\":{\"code\":\"not_found\",\"message\":"
+                        "Result:"
+                        f"{output}")
+
+    @integration_test()
     def test_policy_replace_load_combo_returns_help_screen(self):
         with redirect_stderr(self.capture_stream):
             output = self.invoke_cli(self.cli_auth_params,


test/test_integration_policy.py
```diff
@@ -116,6 +116,13 @@ def test_policy_bad_syntax_raises_error(self):
         self.assertIn("422 Unprocessable Entity {\"error\":{\"code\":\"validation_failed\",\"message\":",
                   str(mock_log.output))
 
+    @integration_test()
+    def test_policy_bad_syntax_raises_error_to_terminal(self):
+        policy = "- ! user bad syntax"
```
eladkug, line 121: can you explain the test its not clear from here ? it will fail on file not exist , no?

test/test_integration_policy.py
```diff
@@ -116,6 +116,14 @@ def test_policy_bad_syntax_raises_error(self):
         self.assertIn("422 Unprocessable Entity {\"error\":{\"code\":\"validation_failed\",\"message\":",
                   str(mock_log.output))
 
+    @integration_test()
+    def test_policy_bad_syntax_raises_error_to_terminal(self):
+        policy_with_bad_syntax = "- ! user bad syntax"
+        with self.assertLogs('', level='DEBUG') as mock_log:
```
eladkug, line 122: we want to verify we see the error without debug

conjur/controller/policy_controller.py
```diff
@@ -35,5 +35,5 @@ def load(self):
         except requests.exceptions.HTTPError as http_error:
             if http_error.response.status_code == 422:
```
eladkug, line 36: If the error is from different type 404? 
I think we need to add this fix in more general place, no?

conjur/controller/policy_controller.py
```diff
@@ -35,5 +35,8 @@ def load(self):
...
                 raise InvalidFormatException(f"{http_error}. The policy was empty or its contents "
-                                             "were not in valid YAML.") from http_error
-            raise
+                                             f"were not in valid YAML."
+                                             f"Error: {http_error.response.text}") from http_error
+            raise requests.exceptions.HTTPError(f"{http_error}."
+                                         f"Error: {http_error.response.text}",
+                                         response=http_error.response) from http_error
```
mbenita-Cyberark, line 42: @szamir1 don't we want to catch the 404 error and raise something like (make sure the resource exist?)

test/test_integration_policy.py
```diff
@@ -116,6 +116,31 @@ def test_policy_bad_syntax_raises_error(self):
...
+                        "Result:"
+                        f"{output}")
+
+    @integration_test()
+    def test_policy_load_bad_branch_recieve_detailed_error(self):
+        with redirect_stderr(self.capture_stream):
+            output = self.invoke_cli(self.cli_auth_params,
+                           ['policy', 'load', '-b', 'wrong-branch', '-f', 
+                           self.environment.path_provider.get_policy_path("resource")], exit_code=1)
+        self.assertIn("{\"error\":{\"code\":\"not_found\",\"message\":", output,
```
mbenita-Cyberark, line 137: does this tests works? looks like you need regex check. but I don't really know what should be the output

conjur/controller/policy_controller.py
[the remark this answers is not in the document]
hzur: I'm editing the error text to include the response in the message so I'll have the explanation why the command failed

test/test_integration_policy.py
```diff
@@ -116,6 +116,31 @@ def test_policy_bad_syntax_raises_error(self):
         self.assertIn("422 Unprocessable Entity {\"error\":{\"code\":\"validation_failed\",\"message\":",
                   str(mock_log.output))
 
+    @integration_test()
+    def test_policy_bad_syntax_error_user_recieve_detailed_error(self):
+        policy_with_bad_syntax = "- ! user bad syntax"
+        with redirect_stderr(self.capture_stream):
+            output = utils.load_policy_from_string(self, policy_with_bad_syntax, exit_code=1)
+        self.assertIn("{\"error\":{\"code\":\"validation_failed\",\"message\":", output,
```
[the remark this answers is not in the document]
hzur: as we used it this way in the over tests, won't it be better to leave it like this?

test/test_integration_policy.py
hzur: yes, apparently it does work, as you can see the pipeline passed and it passes locally

conjur/controller/policy_controller.py
mbenita-Cyberark: i think this is a policy related error handling. So this is the place. agree server should not return 404 on policy not found. weird no?

test/test_integration_policy.py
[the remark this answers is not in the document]
mbenita-Cyberark: weird I dont see we are using `\"` except in `test_policy_bad_syntax_raises_error` which is weird

conjur/controller/policy_controller.py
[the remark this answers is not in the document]
mbenita-Cyberark: Looks like some kind a generic issue that would happen to all our HTTP errors. wouldn't it be better to refer to this use case  in the error handling function `determine_status_code_specific_error_messages`? looks like we need to create a patch to fix our default error handling behavior

test/test_integration_policy.py
mbenita-Cyberark: sorry I thought the error code should be a number

conjur/controller/policy_controller.py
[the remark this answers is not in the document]
hzur: the question is who is responsible to add the error details.
I believe that the responsibility to explain the error should be the responsibility of the controller.
In an ideal world we'll have a dedicated error class for each error type and the controller will construct it

szamir1 (changes requested):

szamir1 (approved):

eladkug (approved):
LGTM

PR #376, opened 2021-12-19
Extend linting check line length to 120

mbenita-Cyberark:
### Desired Outcome

pipeline linting stage will now require less then 120 chars in line instead of 100

### Implemented Changes

- Added pylintrc file with the configuration
- Extended lien 21 in http_wrapper.py to test pipeline

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,2 @@
+[FORMAT]
+max-line-length=120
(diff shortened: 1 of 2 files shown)


.pylintrc
```diff
@@ -0,0 +1,2 @@
+[FORMAT]
+max-line-length=120
```
abrahamko, line 2: add new line

mbenita-Cyberark: Done

abrahamko (approved):
LGTM, thanks for that!

abrahamko (approved):
LGTM

PR #381, opened 2021-12-21
Restore build_publish_container

mbenita-Cyberark:
### Desired Outcome

Restore file that was deleted in the past that is needed for the publish


bin/build_publish_container
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+
+docker build -f Dockerfile.publish \
+             -t conjur-api-python3-publish \
+             .
```
tzheleznyak, line 5: EOF

tzheleznyak (approved):
LGTM

PR #382, opened 2021-12-22
Give permissions to build_publish_containers file

mbenita-Cyberark:
### Desired Outcome

Publish containers will be executed on version change

### Implemented Changes
The problem is that the file has insufficient permissions to be executed.
in this PR we gave it a chmod +x permissions


abrahamko (approved):
LGTM

PR #408, opened 2022-05-09
Add authn-ldap design spec

szh:
### Desired Outcome

Add design document for authn-ldap support in the CLI and API.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-20237](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-20237)

### DIFF
--- a/design/authn-ldap.md
+++ b/design/authn-ldap.md
@@ -1,0 +1,225 @@
+# Solution Design - Authn-LDAP Support in CLI
+
+## Table of Contents
+
+- [Solution Design - Authn-LDAP Support in CLI](#solution-design---authn-ldap-support-in-cli)
+  - [Table of Contents](#table-of-contents)
+  - [Useful Links](#useful-links)
+  - [Overview](#overview)
+    - [Support in previous CLI versions](#support-in-previous-cli-versions)
+    - [Architectural motivation](#architectural-motivation)
+  - [Solution](#solution)
+    - [User Experience](#user-experience)
+      - [Feature Details](#feature-details)
+      - [Out of Scope](#out-of-scope)
+  - [Performance](#performance)
+  - [Backwards Compatibility](#backwards-compatibility)
+  - [Affected Components](#affected-components)
+  - [Test Plan](#test-plan)
+    - [Test Cases](#test-cases)
+      - [Unit Tests](#unit-tests)
+      - [E2E tests](#e2e-tests)
+      - [Security testing](#security-testing)
+  - [Logs](#logs)
+  - [Documentation](#documentation)
+  - [Version update](#version-update)
+  - [Security](#security)
+  - [Audit](#audit)
+  - [Solution Review](#solution-review)
+
+## Useful Links
+
+| Link | Private |
+|------|:-------:|
+| [Conjur Docs: LDAP Authentication - Overview](https://docs.conjur.org/Latest/en/Content/Integrations/ldap/ldap-sync-authn-integration.htm) | No |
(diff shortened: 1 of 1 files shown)


design/authn-ldap.md
```diff
@@ -0,0 +1,236 @@
...
+| Link | Private |
+|------|:-------:|
+| [Conjur Docs: LDAP Authentication - Overview](https://docs.conjur.org/Latest/en/Content/Integrations/ldap/ldap-sync-authn-integration.htm) | No |
+| [Conjur Docs: Configure Developer Environment to Use LDAP Authentication](https://docs.conjur.org/Latest/en/Content/Integrations/ldap/ldap-authn-configure-dev-env.htm) (For Ruby-based) | No |
+| [Pseudo-code for AuthenticationStrategy interface](https://gist.github.com/jvanderhoof/c0457452ec83dbf5c31527edfb8c1dcb) | No |
+
+## Overview
+
+Conjur supports various forms of authentication, including LDAP. This document describes the design and implementation of the LDAP authentication
+support in the new Pytyon-based Conjur CLI.
```
jvanderhoof, line 47:
```suggestion
support in the new Python-based Conjur CLI.
```

szh: Ah, I invented a new language!

design/authn-ldap.md
```diff
@@ -0,0 +1,236 @@
...
+the conjur-api-python library to be constructed in a way that allows us to easily add support for different authentication methods by
+implementing a common `AuthenticationStrategy` interface. Since the CLI and API are now in separate repositories, much of the implementation for
+this design will occur in the API, which will move us closer to a pluggable authentication flow.
+
+## Solution
+
+### User Experience
+
+The `conjur login` command should support LDAP authentication. It should be possible to specify the authentication type using the
+`-t` or `--type` option. The default value should be `authn` (same as current), and it should additionally support `ldap`.
```
jvanderhoof, line 72: I like `-t` as the short flag option. I also like the descriptiveness of the `--authn-type` flag suggested below.

design/authn-ldap.md
```diff
@@ -0,0 +1,236 @@
...
+When using `ldap`, a `--service-id` option should be mandatory.
+
+Other argument names to consider:
+- `--authn-type`
+- `--ldap-service-id`
```
jvanderhoof, line 77: I really like this as an option as well. Having an authenticator specific flag means an end user only needs to set one flag to achieve their outcome instead of two flags (`--type` and `--service-id`). We may want to consider implementing both options if it doesn't increase scope significantly.

szh: I don't think it would increase scope, but having multiple ways to do things will increase complexity and documentation, which we may want to consider.

jvanderhoof:
@szh, how does the CLI (and by extension, the SDK), currently store the auth token after authorization?

szh:
It appears that the SDK itself does not store the API key outside of memory, but the CLI stores it in a `CredentialsProvider`.

design/authn-ldap.md
```diff
@@ -0,0 +1,225 @@
...
+</tr>
+<tr class="odd">
+<td>QA Architect</td>
+<td>Adam Ouamani</td>
+<td><ul>
+<p> </p>
+</ul></td>
+</tr>
+</tbody>
+</table>
```
jvanderhoof, line 225:
```suggestion
Persona | Name | Design Approval
-- | -- | --
Team Leader | John Tuttle |  |
Product Owner | Jane Simon |  |
System Architect | Jason Vanderhoof | [x] |
Security Architect | Andy Tinkham |  |
QA Architect | Adam Ouamani|  |
```
My attempt to approve this design 😄 !

jvanderhoof (approved):
Looks good.  Great work @szh.

PR #412, opened 2022-07-07
Make password prompt specific to authn type

szh:
### Desired Outcome

When running the `conjur login` command without passing the username and password via command line arguments, the CLI prompts the user to enter a username and password. The text for the password prompt is currently: `Enter your password or API key`. This is misleading when using LDAP to login because the API key cannot be used to login via authn-ldap. The prompt text should be accurate for both regular authn (default Conjur authentication) and authn-ldap modes.

### Implemented Changes

- Set the prompt text based on the authentication strategy configured in `conjur init` and saved in the `.conjurrc` file.
- Authn: `Enter your Conjur password or API key`
- LDAP: `Enter your LDAP password`

### Connected Issue/Story

N/A

#### Changelog

- [x] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

- [ ] This PR includes new unit and integration tests to go with the code
  changes, or

#### Documentation

#### Behavior

#### Security

- [x] There are no security aspects to these changes

### DIFF
--- a/conjur/controller/login_controller.py
+++ b/conjur/controller/login_controller.py
@@ -13,6 +13,7 @@ import getpass
 # SDK
 from conjur_api.models import SslVerificationMetadata, SslVerificationMode
 from conjur_api.models import CredentialsData
+from conjur.data_object.authn_types import AuthnTypes
 
 # Internals
 from conjur.util import util_functions
@@ -47,10 +48,11 @@ class LoginController:
         """
         Method that facilitates all method calls in this class
         """
-        self.get_username()
-        self.get_password()
-
         conjurrc = self.load_conjurrc_data()
+
+        self.get_username()
+        self.get_password(conjurrc)
+
         self.get_api_key(conjurrc)
 
         self.login_logic.save(self.credential_data)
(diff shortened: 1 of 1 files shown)


conjur/controller/login_controller.py
```diff
@@ -96,3 +97,12 @@ def get_api_key(self, conjurrc: ConjurrcData):
                                                                      self.credential_data,
                                                                      self.user_password,
                                                                      conjurrc)
+
+    def _get_password_prompt(self, conjurrc: ConjurrcData = None):
```
gl-johnson, line 101: Very small nit, and maybe serves a purpose that I'm not aware of (in which case feel free to ignore) - it seems slightly redundant to have both `get_password` and `_get_password_prompt` providing a default value of `None` to conjurrc when the value is not changed between calls

gl-johnson (approved):
Just a tiny nit, but otherwise LGTM

conjur/controller/login_controller.py
szh: Yeah good point. I fixed that.

PR #413, opened 2022-07-11
Revert conjurrc key names to match Ruby CLI

szh:
### Desired Outcome

Update the Python CLI to use the older names for the keys in the conjurrc file. This means changing conjur_account => account and conjur_url => appliance_url.

### Implemented Changes

When reading or writing .conjurrc:

- `conjur_account` is now `account`
- `conjur_url` is now `appliance_url`

For now we are still supporting the old keys when *reading* the conjurrc in order not to break existing installations of the 7.x CLI. This is deprecated and will be removed in a future release.

This fixes incompatibility between this Python CLI and the previous
Ruby/Docker based CLI, as well as summon-conjur.

The variable names remain the same. Only the key names are changed.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-22955](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-22955)

#### Changelog

- [x] The CHANGELOG has been updated, or
- [ ] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

- [x] This PR includes new unit and integration tests to go with the code
  changes, or

#### Documentation

#### Behavior

#### Security

- [x] There are no security aspects to these changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
-## [7.2.0] - 2022-06-15
+## [7.2.0] - 2022-07-11
+
+### Changed
+- Revert conjurrc key names to match Ruby CLI.
+  [cyberark/cyberark-conjur-cli#413](https://github.com/cyberark/cyberark-conjur-cli/pull/413)
 
 ### Added
 - Support authn-ldap
--- a/conjur/config.py
+++ b/conjur/config.py
@@ -31,8 +31,8 @@ class Config():
     # We intentionally remap some fields to friendlier names
     # Conjurrc field / Config name / Mandatory
     FIELDS = [
-        ('conjur_account', 'account', True),
-        ('conjur_url', 'url', True),
+        ('account', 'account', True),
+        ('appliance_url', 'url', True),
         ('cert_file', 'ca_bundle', True),
         ('authn_type', 'authn_type', False),
     ]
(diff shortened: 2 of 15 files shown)


rpothier (commented):
Code changes look good!
Does this mean we're stuck with the old names forever?
Could we read both the old and the new and only write the new so we upgrade everyone?

szh:
> Code changes look good! Does this mean we're stuck with the old names forever?

@rpothier In planning we pointed out that the Python CLI is not really supported yet (only Community level) so we're ok with some breaking changes, and that we may want to come back around and change these values at some point in the future but for now there's no real reason to change them.

> Could we read both the old and the new and only write the new so we upgrade everyone?

@jtuttle Do you think it's worth adding this in now? Is there a large number of users using the Python CLI?

jtuttle:
@szh Yeah I guess it's pretty easy to read both so let's go ahead and do that. We can steer people towards one or the other via documentation and deprecation notices.

szh:
> @szh Yeah I guess it's pretty easy to read both so let's go ahead and do that. We can steer people towards one or the other via documentation and deprecation notices.

OK great. I've added that functionality and updated the PR description.

CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
-## [7.2.0] - 2022-06-15
+## [7.2.0] - 2022-07-11
+
+### Changed
+- Revert conjurrc key names to match Ruby CLI. **This is a breaking change.**
```
rpothier, line 12: Is this still considered a braking change?

szh: Good question. I thought maybe in case a user has custom code that depends on reading the conjur_account and conjur_url keys, but I think you're right that it isn't. More of a deprecation.

rpothier (approved):
LGTM!

PR #423, opened 2022-08-23
Updates 'init' command's cert_file property

juniortaeza:
### Desired Outcome

- Added fixes to the 'init' command to omit empty string paths for Ruby CLI compatibility

### Implemented Changes

- Updated files to remove empty string declarations

### Connected Issue/Story

https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-24956

### DIFF
--- a/conjur/controller/init_controller.py
+++ b/conjur/controller/init_controller.py
@@ -79,7 +79,7 @@ class InitController:
             fetched_certificate = self._get_server_certificate(formatted_conjur_url)
             self._write_certificate(fetched_certificate)
         elif mode in [SslVerificationMode.INSECURE, SslVerificationMode.TRUST_STORE]:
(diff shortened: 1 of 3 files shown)


szh:
Do we have integration tests that check that the cert_file will actually not be in the config file in this case?

gl-johnson (commented):
Following Shlomo's suggestion, have a look at some of the tests in [/test/test_integration_configurations.py](https://github.com/cyberark/cyberark-conjur-cli/blob/99cf817c1d80b3833601fffa642d4ad1e37c6798/test/test_integration_configurations.py#L126). 

It would be worthwhile to add a similar test where we invoke the CLI init command with neither cert-related flag and verify that conjurrc contains `cert_file: null` (as opposed to an empty string)

PR #424, opened 2022-08-24
Solution Design - Authn-OIDC/MFA Support

szh:
### Desired Outcome

OIDC support was recently added to the Conjur UI. See the [supporting documentation](https://ca-il-confluence.il.cyber-ark.com/display/rndp/MFA+for+Conjur+UI+-+PI+2022.3) (internal link) for this effort for more details. The goal of this effort will be to add OIDC support to the new (Python) Conjur CLI such that a user can login to Conjur through Okta.

The AWS CLI allows OIDC / MFA authentication by popping open the system's browser. We should explore this approach during design but also keep in mind that this won't work on systems without a GUI (i.e. when running on a Docker container).

In the future, we will also want to support Microsoft ADFS and CyberArk Identity (in addition to Okta) so we should try to design Okta support in a flexible enough way to support additional vendors. Since they all use the OIDC standard this should hopefully be fairly straightforward.

### Implemented Changes

Created Solution Design document and Jira stories.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-24839](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-24839)


design/authn-oidc.md
```diff
@@ -0,0 +1,121 @@
...
+| `--authn-type` / `-t` | `authn`, `ldap`, `oidc` | ??? | The authentication method to use when connecting to Conjur |
+| `--service-id` | `<service_id>` | N/A | The service_id of the OIDC service to use |
+
+The `conjur login` command should either:
+
+1. Open a browser to the OIDC URL configured for the SSO provider
+2. Print the URL (and/or a QR code) to be copied to a browser when running on a text-only machine
+
+The user must then authenticate to the SSO provider and receive a JWT which will be stored in the user's keychain or in the .netrc file.
+*TODO: How will the CLI get the JWT token from the browser?*
```
gl-johnson, line 73: > *TODO: How will the CLI get the JWT token from the browser?*

This is the million dollar question to me. Generally the OIDC provider would make a callback containing the authorization code via the `redirect_uri` that is provided in the initial authn request. How will it work if we aren't running a webserver somewhere that we can retrieve the code from?

It looks like there are some flows that don't require redirects, i.e. [the resource owner password flow](https://developer.okta.com/docs/concepts/oauth-openid/#resource-owner-password-flow) so I'm assuming we are limited to those

design/authn-oidc.md
```diff
@@ -0,0 +1,121 @@
...
+#### Feature Details
+
+##### In the CLI repository
+
+The CLI needs to support the new CLI argument values in the `init` command, and it will need to pass them to the conjur-api-python library.
+Regarding the `login` step, this will be implemented very differently than the other authentication strategies (see details
+[below](#in-the-api-repository)).
+While this will mostly be contained in the API repository, there will need to be changes made to the login controller and logic
+in the CLI repository. For instance, when using authn-oidc, we won't be prompting the user for a username and password as this will
+be determined by the SSO provider.
```
gl-johnson, line 86: Plus any changes to .netrc (or keyring) which I assume is where the token(s) will be stored.

design/authn-oidc.md
```diff
@@ -0,0 +1,121 @@
...
+
+##### In the API repository
+
+Create a new implementation of `AuthenticationStrategyInterface` called `OidcAuthenticationStrategy` which will handle the
+authentication logic. In it's `login()` method, it should reach out to the Conjur API and retrieve the necessary redirect URL
+for the specified OIDC authenticator, and then open a browser (or print a link) to the redirect URL. The API code must retrieve
+the JWT from the browser and save it in the `CredentialsProviderInterface`.
```
gl-johnson, line 93: I'm curious how this will work if the client needs to make a [POST request to the token endpoint](https://auth0.com/docs/get-started/authentication-and-authorization-flow/call-your-api-using-resource-owner-password-flow#example-post-to-token-url) (with some data)?

gl-johnson (commented):
Looking good so far. Highlighted my biggest unknowns.

design/authn-oidc.md
szh: Good point. One option I think I've seen apps use is to start a web server on localhost and set that as the redirect URL. That way we can get the JWT and then pass it along to the Conjur server and we can use the [Authorization Code with PKCE flow](https://developer.okta.com/docs/guides/implement-grant-type/authcodepkce/main/) which is recommended. I'll add this to the document but I'm definitely interested in other approaches.

design/authn-oidc.md
szh: I don't think this will require a change in the CLI project because the JWT will just replace the API token in the `CredentialsData` class.

design/authn-oidc.md
szh: Looking around online, it seems that popping open a browser is the way to go. See for example [this tutorial](https://medium.com/@balaajanthan/openid-flow-from-a-cli-ac45de876ead) and [this sample code](https://github.com/puiterwijk/python-openidc-client/blob/master/openidc_client/__init__.py#L499). If we go this route, OIDC will not be supported on remote servers.

jvanderhoof:
@szh, it came to light this morning that we need to make `state` and likely `nonce` dynamically generated per authentication request. For the UI, this is fairly straight forward as we can take advantage of sessions. I'm concerned what this might look like in the CLI. 

Beyond this question, I'm wondering:
- How will the CLI handle the browser redirect after successful authentication?
- How does the CLI get the `code` and `state` values from the redirect?

Could you please include a workflow diagram that shows how the data will move between the CLI, browser, CLI, and Conjur?

szh:
@jvanderhoof I'll work on a diagram. In the meantime, the answer to these questions:

> How will the CLI handle the browser redirect after successful authentication?
  How does the CLI get the code and state values from the redirect?

is through the ad-hoc HTTP server that will be launched on localhost. See this [code example](https://github.com/puiterwijk/python-openidc-client/blob/master/openidc_client/__init__.py#L446) for an illustration.

design/authn-oidc.md
```diff
@@ -0,0 +1,250 @@
...
+## Solution
+
+### User Experience
+
+The `conjur init` command should support OIDC authentication as an authentication type. It should be possible to specify
+the authentication type `oidc` using the existing `-t` or `--authn-type` option.
+When using `oidc`, a `--service-id` option should be mandatory.
+
+*Open question*: Currently, if the `--service-id` option is specified, then `--authn-type` defaults to `ldap`. Should we make this required
+without a default value instead?
```
szh, line 59: @jSimonCA Can we get your opinion on this as PO?

design/authn-oidc.md
```diff
@@ -0,0 +1,257 @@
...
+Then when executing any CLI commands, the CLI will retrieve the stored JWT and pass it to Conjur's OIDC authentication endpoint and receive
+a short-lived accesss token, as with all other authentication strategies.
+
+*Open question*: Can we revoke the JWT when `conjur logout` is called?
+
+#### Feature Details
+
+##### Diagram
+
+![Diagram](assets/authn-oidc.png)
```
gl-johnson, line 79: I'm still a bit confused by our usage of `redirect_uri` in this context.  Usually in OAuth/OIDC the redirect_uri refers to the endpoint which the OIDC provider redirects the authorization code and state/nonce data to **after a successful authentication attempt**. Here it seems like we are calling the Okta authorization endpoint that gets opened on the client browser a `redirect_uri`.  Later in the diagram (and this document) we are referring to the redirect_uri that Okta uses as a callback_url.

In my mind, the term `redirect_uri should be reserved for the required parameter sent to Okta in the authentication request [per the ODIC spec](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest). 

Example (assuming ad-hoc HTTP server):
redirect_uri = https://localhost:{port}/handleOidcRedirect

Authorization endpoint = https://cyberark.okta.com?client_id=<client_id>&redirect_uri=<redirect_uri>&response_type=code&scope=openid%2fprofile%2femail&state=<state>&nonce=<nonce>

Maybe @jvanderhoof can weigh in on this since I think this diagram is based on the ConjurUI design.

szh: Ah sorry - I updated the language in the markdown but forgot to in the diagram. I'll fix this now.

szh: It's true that it's called "redirect_uri" in the Conjur UI design doc but that's probably because from the perspective of the UI which is a web app, it's the URL that the user will be *redirected* to. I'm going to try to make it a bit clearer in the diagram here.

szh: @gl-johnson Diagram updated. Please let me know if that's clearer.

gl-johnson: Yep - thanks for addressing

design/authn-oidc.md
jSimonCA: Reviewed the solution design and workflow/customer experience looks good. Few questions/feedback from my end:
1. How is token refresh scenario handled? This is something that is getting redefined with 12.8 for MFA UI implementation so maybe we should align user experience for CLI also?
2. Please ensure that state, nonce, and code_challenge (PKCE - Proof Key for Code Exchange) are included as per security findings that came up last week. I see Jason has already mentioned it as part of his feedback as well.

